PR #51, opened 2014-08-11
Support materialized views

calebthompson:
* create, drop, and update materialized views in migrations
* dump materialized views into schema statements
* automatically generate code for model and migration in migrations

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+addons:
+  postgresql: "9.3"
 before_install:
   - "echo '--colour' > ~/.rspec"
   - "echo 'gem: --no-document' > ~/.gemrc"
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,9 @@
 # News
 
-The noteworthy changes for each Clearance version are included here. For a
-complete changelog, see the git history for each version via the version links.
+The noteworthy changes for each Scenic version are included here. For a complete
+changelog, see the [CHANGELOG] for each version via the version links.
+
+[CHANGELOG]: https://github.com/thoughtbot/scenic/commits/master
 
 ## [0.3.0] - January 23, 2015
 
--- a/lib/generators/scenic/view/view_generator.rb
+++ b/lib/generators/scenic/view/view_generator.rb
@@ -1,10 +1,12 @@
 require "rails/generators"
 require "rails/generators/active_record"
+require "generators/scenic/materializable"
 
 module Scenic
   module Generators
     class ViewGenerator < Rails::Generators::NamedBase
       include Rails::Generators::Migration
+      include Scenic::Generators::Materializable
       source_root File.expand_path("../templates", __FILE__)
 
       def create_views_directory
@@ -25,12 +27,12 @@ module Scenic
         if creating_new_view? || destroying_initial_view?
           migration_template(
             "db/migrate/create_view.erb",
-            "db/migrate/create_#{plural_file_name}.rb"
+            "db/migrate/create_#{plural_file_name}.rb",
           )
         else
           migration_template(
             "db/migrate/update_view.erb",
-            "db/migrate/update_#{plural_file_name}_to_version_#{version}.rb"
+            "db/migrate/update_#{plural_file_name}_to_version_#{version}.rb",
           )
         end
       end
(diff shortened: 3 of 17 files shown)


guiman:
Hi @calebthompson @derekprior I would really much like this feature in. Any way I could help move this forward?

calebthompson:
@derekprior this is ready I think.

NEWS.md
```diff
@@ -44,3 +51,7 @@ definition files.
 In short, go add a view to your app.
 
 [0.1.0]: https://github.com/thoughtbot/scenic/compare/8599daa132880cd6c07efb0395c0fb023b171f47...v0.1.0
+=======
```
derekprior: git conflicts?

lib/generators/scenic/view/view_generator.rb
```diff
@@ -7,6 +7,20 @@ class ViewGenerator < Rails::Generators::NamedBase
...
+        @name_with_arguments = name
+        self.name = args.shift
+        assign_names!(name)
+
+        args.each do |arg|
+          setter = "#{arg}="
+          if respond_to?(setter)
+            send(setter, true)
+          end
+        end
```
derekprior: This is entirely duplicated in the generator above. If we're going to handle arguments this way, we should extract a parser object for this.

derekprior:
I feel less strongly about the generator syntax suggestion I made months ago, but if you're happy with `scenic:view people:materalized` I'm find with it. I briefly entertained `scenic:materialzed_view people` but I don't like the corollary `scenic:materialized_model people`.

I also think that since this is a pretty major feature that neigher of us has used, we should merge this as part of `1.0.0.beta1` and see if we can get some brave souls to give it a shot in real use with upgrades, refreshes, and drops.

calebthompson:
I had a commit that changed the generator to `scenic:view people --materialized` using the `class_option` macro, but it broke some tests and I didn't want to fight with it.

derekprior:
If we merge #81 (and I think it's definitely the right thing to do), we'll probably need to figure out the argument syntax since I don't think the rails model generator will be able to understand `people:materialized`.

calebthompson:
I have WIP to us class_option. It's fighting me and I'm busy with other things so it's not a priority for me right now.

nicolai86:
@calebthompson what's missing to land this? any todo list I could jump in and help you out with to finish this PR?

derekprior:
Just thought of one potential gotcha we should discuss: Materialized views can have indexes. In fact, I'd say that most probably will. Our update strategy is to drop and recreate the view. This would eliminate all of the indexes.

Do you think we should disallow `update_view` for materialized views and make people be more explicit with drop and create? Is there something else we can do to preserve indexes?

calebthompson:
I just pushed up the latest commit, which breaks the build. @nicolai86 you want to take a whack at fixing that (I suspect it's a test setup issue, but I'm not sure), I'd be happy to merge a PR to this branch, and rebase to give credit for the whole change.

calebthompson:
Good point, @derekprior. We've got knowledge, potentially, of all the indices for the view. I wonder if we could recreate them as well when we rebuild the table.

calebthompson:
Oh and it might make sense to add a class method to refresh the view.

derekprior:
@calebthompson This is rebased on master and all green. Only functional change I made is contained in the latest commit and, for now, disallows updating a materialized view.

calebthompson:
lgtm

.travis.yml
```diff
@@ -1,3 +1,5 @@
+addons:
+  postgresql: "9.3"
```
calebthompson: 9.4 now maybe?

NEWS.md
```diff
@@ -1,7 +1,9 @@
 # News
 
-The noteworthy changes for each Clearance version are included here. For a
-complete changelog, see the git history for each version via the version links.
+The noteworthy changes for each Scenic version are included here. For a complete
+changelog, see the [CHANGELOG] for each version via the version links.
+
+[CHANGELOG]: https://github.com/thoughtbot/scenic/commits/master
```
calebthompson: tehee hi @olivierlacan

.travis.yml
derekprior: Materialized view support was added in 9.3, so it shouldn't matter.

Issue #52, opened 2014-08-19
Error if db/views does not exist


calebthompson:
```
/Users/caleb/.rvm/gems/ruby-2.1.2@team/gems/scenic-0.2.0/lib/generators/scenic/view/view_generator.rb:35:in `open': No such file or directory @ dir_initialize - /Users/caleb/code/team/db/views (Errno::ENOENT)
        from /Users/caleb/.rvm/gems/ruby-2.1.2@team/gems/scenic-0.2.0/lib/generators/scenic/view/view_generator.rb:35:in `entries'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@team/gems/scenic-0.2.0/lib/generators/scenic/view/view_generator.rb:35:in `previous_version'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@team/gems/scenic-0.2.0/lib/generators/scenic/view/view_generator.rb:41:in `version'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@team/gems/scenic-0.2.0/lib/generators/scenic/view/view_generator.rb:64:in `definition'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@team/gems/scenic-0.2.0/lib/generators/scenic/view/view_generator.rb:11:in `create_view_definition'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@global/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@global/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@global/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `block in invoke_all'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@global/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `each'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@global/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `map'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@global/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `invoke_all'
        from /Users/caleb/.rvm/gems/ruby-2.1.2@global/gems/thor-0.19.1/lib/thor/group.rb:234:in `dispatch'
[...]

PR #61, opened 2015-01-05
Handle non-existing db/views directory scenario for generators.

DamirSvrtan:


calebthompson:
Should solve #52 when merged.

lib/generators/scenic/view/view_generator.rb
```diff
@@ -8,6 +8,7 @@ class ViewGenerator < Rails::Generators::NamedBase
       source_root File.expand_path("../templates", __FILE__)
 
       def create_view_definition
+        empty_directory views_path unless views_path.exist?
```
calebthompson: Please no trailing conditionals.

DamirSvrtan:
Removed trailing conditional

lib/generators/scenic/view/view_generator.rb
```diff
@@ -8,6 +8,10 @@ class ViewGenerator < Rails::Generators::NamedBase
       source_root File.expand_path("../templates", __FILE__)
 
       def create_view_definition
+        unless views_path.exist?
```
derekprior: should this be its own public method? The public methods in thor tasks are the steps in a process. This could be "create_views_directory" which I think would describe the process better.

DamirSvrtan: Yes, I thought to my self that there should be a better name. I like create_views_directory or create_db_views_directory.

DamirSvrtan: Actually, I would rename the existing one to 'views_directory_path' and create another one called 'create_views_directory'.

DamirSvrtan:
Should I squash the commits?

derekprior:
Looks good, yes.

DamirSvrtan:
Squashed

PR #63, opened 2015-01-12
Update Travis Configuration

derekprior:
* Use the latest version of Ruby 2.1
* Add Ruby 2.2.0 to the testing matrix
* Enable the Travis container infrastructure (sudo: false)
* Disable bundler cache. We want Travis to resolve and install
 dependencies each time, as we want to know about issues that pop up
 with newer dependency versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,16 @@ before_install:
   - git config --global user.name 'Travis CI'
   - git config --global user.email 'travis-ci@example.com'
 before_script:
-  - pushd spec/dummy && bundle exec rake db:create && popd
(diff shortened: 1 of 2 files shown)


calebthompson:
Yup, but you've failures.

derekprior:
Sure do. Not sure why. Going to try it without the container settings.

derekprior:
... Interesting...

calebthompson:
Yes it is. Our builds aren't super long though, so I'm happy to take what we can get here.

derekprior:
Merging without the container infrastructure changes if this passes.

PR #65, opened 2015-01-15
DROP VIEW IF EXISTS view_name

robertomiranda:
this is useful if your migrating from another solution

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -13,8 +13,11 @@ module Scenic
         execute "CREATE VIEW #{name} AS #{sql_definition};"
       end
 
-      def self.drop_view(name)
-        execute "DROP VIEW #{name};"
+      def self.drop_view(name, if_exists = false)
+        sql = "DROP VIEW"
+        sql += " IF EXISTS" if if_exists
+        sql += " #{name};"
+        execute sql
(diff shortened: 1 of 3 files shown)


derekprior:
Can you explain why this is helpful in migrating from another solution?

calebthompson:
If there's an existing view that you're migrating from, does it make sense to store that in a [name}_v1.sql and start your first scenic-managed view at v2?

In that case, you'd change the drop to an update.

derekprior:
Closing due to lack of activity. If you feel we should reconsider this, please let us know.

ericboehs:
I think it's useful when migrating from another solution.

* We currently have db views in our production db. They're used for analytics and our dev envs don't have them.
* We decided to use Scenic and bring them into our version control because we can't drop columns on tables that the views reference.
* When bringing the existing views into Scenic, we wrote a migration to drop all the views. And then wrote migrations for all the views. We ran these all at once since our db essentially always had the views in place.
* Since the dev envs didn't have the views, the migrations failed and each dev had to manually skip the drop views migration.

It wasn't a huge problem but if I would have had this feature all the devs on our team wouldn't have had extra work when we added Scenic.

I'd imagine there are other use case for this but since Rails has if exists checks for tables and columns, it would make sense to have them views.

PR #69, opened 2015-01-28
Support Ruby 2.0.0

robertomiranda:
While Loading the rails console we're getting this error
```scenic.rb:12:in `load': private method `include' called for ActiveRecord::ConnectionAdapters::AbstractAdapter:Class (NoMethodError)````

We fixed the calling the method ìnclude`with send, what do you think guys?

### DIFF
--- a/lib/scenic.rb
+++ b/lib/scenic.rb
@@ -9,9 +9,9 @@ require "scenic/view"
 
 module Scenic
   def self.load
-    ActiveRecord::ConnectionAdapters::AbstractAdapter.include Scenic::Statements
-    ActiveRecord::Migration::CommandRecorder.include Scenic::CommandRecorder
-    ActiveRecord::SchemaDumper.include Scenic::SchemaDumper
+    ActiveRecord::ConnectionAdapters::AbstractAdapter.send(:include, Scenic::Statements)
+    ActiveRecord::Migration::CommandRecorder.send(:include, Scenic::CommandRecorder)
+    ActiveRecord::SchemaDumper.send(:include, Scenic::SchemaDumper)
   end
 
   def self.database

### CODE REVIEW

**Comment by derekprior** [comment]
This looks good to me, but I don't think this has anything to do with Ruby 2.0.0. This should be an issue in all Rubies though.

**Comment by robertomiranda** [comment]
Not sure, is really weird, we switched to 2.1 and everything works well without this change

**Comment by robertomiranda** [comment]
@derekprior look

`Module#include` is a private method in ruby 2.0.0
```bash
2.0.0-p481 :001 > Module.public_methods.sort.grep(/include/)
 => [:include?, :included_modules] 
```

but in 2.1 or 2.2 not
```bash
 Module.public_methods.sort.grep(/include/)
 => [:include, :include?, :included_modules] 
```

```bash
Module.public_methods.sort.grep(/include/)
 => [:include, :include?, :included_modules] 
```

robertomiranda:
FYI https://bugs.ruby-lang.org/issues/8846

calebthompson:
Blargh. Yeah, this should be merged, then.

Can you add `2.0` to .travis.yml (and drop patch versions for other rubies)?

derekprior:
I'm not certain we want to support 2.0. Is there a compelling reason to do
so? What if we want to use a required keyword argument somewhere?

calebthompson:
I'd love not to, but my general stance is that we support whatever Heroku supports for Ruby and whatever Rails supports for Rails, which puts us at `2.0, 2.1, and 2.2` and `4.2`.

derekprior:
Having to use `send` rather than just calling a public method seems like a decent enough reason to just say we're going to use 2.1+ but I don't really feel strongly enough to object any further.

calebthompson:
Yeah I mostly agree. Closing. Sounds like @robertomiranda isn't having the issue anymore since you've upgraded to 2.1?

PR #75, opened 2015-09-29
Improve formatting of dumped view

derekprior:
* Blank line should go above the view to separate it from content above
  and still guarantee a blank line between views
* `stip_heredoc` wasn't working as expected. Postgres does a reasonable
  job of formatting the query, we just want to be sure it is readable in
  the schema. `indent` helps with that.

*Before:*
janky

*After:*
better

### DIFF
--- a/lib/scenic/view.rb
+++ b/lib/scenic/view.rb
@@ -14,11 +14,11 @@ module Scenic
     end
 
     def to_schema
-      <<-DEFINITION.strip_heredoc
-        create_view :#{name}, sql_definition:<<-\SQL
-          #{definition}
-        SQL
+      <<-DEFINITION
 
+  create_view :#{name}, sql_definition: <<-\SQL
+    #{definition.indent(2)}
+  SQL
       DEFINITION
     end
   end


calebthompson:
Can we see the difference in output? Maybe in the commit message?

calebthompson:
and/or a spec

calebthompson:
> This is not the easiest formatting I could find

What was?

derekprior:
@calebthompson Ready for another look. Update implementation and commit (now with images). No tests for the formatting specifically. I don't think it'd be super useful to test it. We can set it and hopefully forget it.

calebthompson:
Starship Enterprise it

calebthompson:
oh wait

calebthompson:
That indentation is still wonky on the first line.

derekprior:
One more time... with FEELING!

terminal_ _tmux_ _tat_ _231x62

lib/scenic/view.rb
```diff
@@ -14,11 +14,11 @@ def ==(other)
...
 
     def to_schema
-      <<-DEFINITION.strip_heredoc
-        create_view :#{name}, sql_definition:<<-\SQL
-          #{definition}
-        SQL
+      <<-DEFINITION
 
+  create_view :#{name}, sql_definition: <<-\SQL
+    #{definition.indent(2)}
```
calebthompson: This line needs to be dedented to 0 column for the `indent` to have the desired effect.

calebthompson: I'd expect `SELECT` and `UNION` to be indented to the same level, making them cocolumnar with `create_view`. I'll leave it up to you whether we should then bump up the `indent` level to 4 to appear as a "body" to the call.

derekprior: That appears to work for some views but not others. Postgres does some weird formatting, evidently. I tried various configurations. The `indent(2)` there *does* have the desired effect.

derekprior: All this to say: It's not perfect, but it's much better.

calebthompson:
:+1:

PR #76, opened 2015-10-04
Fix failing tests with RSpec 3.3

derekprior:
There is no released version of ammeter that works with the latest
version of RSpec for libraries not using fixtures. I've asked for a
release of ammeter, but in the meantime, we need to pull latest from
git. See: https://github.com/alexrothenberg/ammeter/pull/50

### DIFF
--- a/scenic.gemspec
+++ b/scenic.gemspec
@@ -23,10 +23,10 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '>= 1.5'
   spec.add_development_dependency 'database_cleaner'
   spec.add_development_dependency 'rake'
-  spec.add_development_dependency 'rspec', '~> 3.0'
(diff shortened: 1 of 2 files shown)


calebthompson:
Does this need to be merged? I'd almost rather have breaking build for a little while as we don't have a high rate of change otherwise.

That change was made in Ammeter four months ago. Is this a dependency we want to continue to have?

derekprior:
We've had a breaking build for 4 months. I know of no replacements for ammeter and I think energy is better spent on encouraging better stewardship than replacing it.

If this were a runtime dependency I'd be more concerned.

calebthompson:
Okay. Merge if you want to, maybe just create an issue to roll it back.

derekprior:
Updated to use latest ammeter release. Assuming this passes, will merge.

PR #80, opened 2015-11-06
Document issues with ActiveRecord find

### LINKED ISSUE #77: Document How Postgres can't infer a view's primary key
```
@calebthompson: @derek We’re using Scenic. While inspecting with pg, the view has an `id` column, but `TheViewModel.first.id == nil` and `TheViewModel.find(1)` complains about not having a primary key

calebthompson [14:58] 
`self.primary_key = :id`

sean [14:58] 
Setting `self.primary_key = "id"` resolved the issue. Am I doing something wrong

calebthompson [14:58] 
Nope.

calebthompson [14:58]
Views can’t have primary keys in postgres.

calebthompson [14:58]
We could probably benefit from having that in the README.
```

derekprior:
Closes #77.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,18 @@ class Search < ActiveRecord::Base
 end
 ```
 
+## When I query that model with `find` I get an error. What gives?
+
+Your view cannot have a primary key, but ActiveRecord's `find` method expects to
+query based on one. You can use `find_by!` or you can explicitly set the primary
+key column on your model like so:
+
+```ruby
+class People < ActiveRecord::Base
+  self.primary_key = :id
+end
+```
+
 ## Can you make this easier?
 
 Sure thing. How about some generators?


README.md
```diff
@@ -79,6 +79,18 @@ class Search < ActiveRecord::Base
...
 
+## When I query that model with `find` I get an error. What gives?
+
+Your view cannot have a primary key, but ActiveRecord's `find` method expects to
+query based on one. You can use `find_by!` or you can explicitly set the primary
+key column on your model like so:
+
+```ruby
+class People < ActiveRecord::Base
+  primary_key :id
```
calebthompson: I think it's `self.primary_key = :id`.

http://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/PrimaryKey/ClassMethods.html

PR #81, opened 2015-11-07
Delegate to rails model generator

derekprior:
The rails model generator supports lots of things our simplistic
`scenic:model` generator does not. Specifically, the rails model
generator can handle namespaced models by generating the right class
name and a module that sets the table name prefix properly for the
entire namespace.


derekprior:
By using this generator, we also get anything else that hooks into it, like test generators.

calebthompson:
This makes a potential improvement around setting `readonly`, `primary_key`, etc. slightly more difficult, but we can cross that bridge when we come to it. I imagine there's tooling to add to a file for generators.

:+1:

derekprior:
Yeah, we can use `inject_into_file`

PR #82, opened 2015-11-07
Allow configuration of database adapter

derekprior:
At this point, we don't have any plans to offer different in-tree
database adapters, but this minimal configuration should allow other
developers to write them and plug them in without monkey-patching
Scenic.

### DIFF
--- a/lib/scenic/configuration.rb
+++ b/lib/scenic/configuration.rb
@@ -1,0 +1,34 @@
+module Scenic
+  class Configuration
+    # The Scenic database adapter class to use when executing SQL.
+    # Defualts to [Scenic::Adapters::Postgres]
+    # @return Scenic adapter
+    attr_accessor :database
+
+    def initialize
+      @database = Scenic::Adapters::Postgres
+    end
+  end
+
+  # @return [Scenic::Configuration] Scenic's current configuration
+  def self.configuration
+    @configuration ||= Configuration.new
+  end
+
+  # Set Scenic's configuration
+  # @param config [Scenic::Configuration]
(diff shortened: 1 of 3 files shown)


lib/scenic/configuration.rb
```diff
@@ -0,0 +1,34 @@
+module Scenic
+  class Configuration
```
calebthompson: What do you think of using Rails' custom configuration (example: https://github.com/dhh/custom_configuration) instead of building up our own class?

calebthompson: Something like `Rails.application.config.scenic.database`

derekprior: Hmm. I forgot about that being a thing. Do we want to drop 4.1 for it? Is it being at least somewhat widely adopted?

derekprior: It looks like it has to be `Rails.application.config.x.scenic.database` (x for extension, I guess).

I'm not sure how I feel about reaching into the Rails.application.config store versus having developers read from that store to initialize our configuration. The latter seems less magical, but I'm willing to yield to convention here.

derekprior: I looked more into this and I can't find any nstances of gems relying on this configuration store directly. It very much seems like it's intended for applications to store data they pass off to other bits of code, not for gems to hook into.

The kicker for me is that our custom configuration provides typo protection while the rails custom config is essentially a Hashie::Mash that will respond to anything.

derekprior:
Merged, but I'm still willing to revisit this.

PR #84, opened 2015-11-11
Remove use of alias_method_chain

derekprior:
Rails 5 deprecates `alias_method_chain` in favor of `Module.prepend`
which is provided in Ruby 2.


calebthompson:
lgtm

PR #85, opened 2015-11-13
Fix materialized view issues

derekprior:
We were lacking a smoke test for materialized view functionality. Adding it
revealed a couple of issues that needed to be fixed:

1. We never implemented `refresh_materialized_view`.

2. We could not successfully run the `rails destroy scenic:view` with the
   materialized option.

### DIFF
--- a/spec/smoke
+++ b/spec/smoke
@@ -83,6 +83,31 @@ main() {
   [[ ! -f app/models/search.rb ]] || exit 1
   [[ -z "$(find db/migrate -maxdepth 1 -name "*create_searches.rb" -print -quit)" ]] || exit 1
   echo "[success]"
+
+  echo "rails generate scenic:model search (materialized)"
+  rails generate scenic:model search --materialized
+  [[ -f db/views/searches_v01.sql ]] || exit 1
+  [[ -f app/models/search.rb ]] || exit 1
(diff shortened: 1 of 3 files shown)


calebthompson:
HAHAHA

calebthompson:
:(

spec/smoke
```diff
@@ -3,87 +3,113 @@
...
+
+  echo "rake db:rollback"
+  rake db:rollback
+  echo "[success]"
+
+  echo "rails destroy scenic:view search --materialized"
+  rails destroy scenic:model search --materialized
+  [[ ! -f app/models/search.rb ]] || exit 1
+  [[ ! -f db/views/searches_v01.sql ]] || exit 1
+  [[ ! -f app/models/search.rb ]] || exit 1
```
calebthompson: This line duplicates 106

calebthompson:
LGTM, though I'd prefer we didn't switch away from tabs.

derekprior:
Bash style guides I can find are either indifferent on the matter or use spaces. Why do you want to use tabs?

calebthompson:
Most of my shell style opinions are based on Git's CodingGuidelines.

https://github.com/git/git/blob/master/Documentation/CodingGuidelines#L43

derekprior:
I'd like to move that to ruby in any event. I think it will enable more people to contribute and I'll certainly be more comfortable with it, anyway.

PR #87, opened 2015-11-14
Extract an abstract database adapter

derekprior:
`Scenic::Adapters::Abstract` lays out the interface for Scenic database
adapters. It contains template methods for all methods required of a
Scenic adapter with actual implementations for `create_view` and
`drop_view` which are identical for basic use cases across all
relational databases I've encountered.

Using the abstract adapter as a base, for instance, a hypothetical MySQL
adapter would need only to implement the `views` method that retrieves
all views and their associated schema for dumping to the `schema.rb`
file.

As part of this extraction, I converted to using instances for the
adapters rather than class methods. This feels better and wasn't a good
deal of work. I'd prefer instance based APIs rather than class-based
APIs.

Documentation was added for all adapter methods.

### DIFF
--- a/lib/scenic/adapters/abstract.rb
+++ b/lib/scenic/adapters/abstract.rb
@@ -1,0 +1,82 @@
+module Scenic
+  module Adapters
+    # Abstract base class for Scenic database adapters.
+    class Abstract
+      # Returns an array of views in the datbase as [Scenic::View] objects.
+      #
+      # This collection of views is used by the [Scenic::SchemaDumper] to
+      # populate the `schema.rb` file.
+      #
+      # This method has no default implementation and must be implemented by
+      # all adapters.
+      #
+      # @return [Array<Scenic::View>]
+      def views
+        raise NotImplementedError
+      end
+
+      # Creates a view in the database.
+      #
+      # This method has a default implementation that should work for all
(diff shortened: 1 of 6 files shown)


derekprior:
@calebthompson I'm not sure about extracting the abstract adapter bit. Does that provide a bit too much of a contract we'd be beholden to? For instance, right off the bat, I thought about all the various options you can pass to `CREATE VIEW`, `CREATE_MATERIALIZED_VIEW` or `REFRESH_MATERIALIZED_VIEW` that our API provides no affordances for. It's almost like we'd want the Schema statements to accept arbitrary options and pass through anything unidentifed as a third `options` parameter for adapters to do with what they'd like. If we want to do that now, I think this abstract adapter move is something we can do now. If we want to consider out options more, I think we wait.

I *do* really like the move to instances over classes, and the documentation so I'd like to do that part regardless.

calebthompson:
Agreed, there's some good stuff here. I do think ts probably premature to do an extraction. If rather see us work with someone interested in writing and maintaining a Scenic adapter for XSQL so that we are sure the API is what we need.

derekprior:
Going to nix the abstract adapter for now, and merge the isntance changes and docs.

PR #89, opened 2015-11-20
Isolate Postgres view representation

derekprior:
Previously, we constructed a `View` object by passing the Postgres
database result directly to the `View` constructor. This leaks
Postgres-specific information.

This change updates the postgres adapter so it constructs the `View`
object by parsing the database result itself. This will make `View` more
easily used by additional adapters.


calebthompson:
:100:

PR #90, opened 2015-11-20
Document public API

derekprior:
* Added documentation where it was missing.
* Updated documentation to use yard formatting throughout.
* Marked internals as private (`SchemaDumper`, `CommandRecorder`, etc)

I did opt to document `Adapters::Postgres` and `View`. They are not
objects we expect end users to interact with, but the documentation will
be useful for anyone who wants to create an adapter.

Those objects are tagged with `@api extension` so we could filter them
from the default documentation with and update to the `.yardopts` file
if needed.

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -1,5 +1,19 @@
 module Scenic
+  # Scenic database adapters.
+  #
+  # Scenic ships with a Postgres adapter only but can be extended with
+  # additional adapters. The {Adapters::Postgres} adapter provides the
+  # interface.
   module Adapters
+    # An adapter for managing Postgres views.
+    #
+    # **This object is used internally by adapters and the schema dumper and is
+    # not intended to be used by application code. It is documented here for
+    # use by adapter gems.**
+    #
+    # For methods usable in migrations see {Statements}.
+    #
+    # @api extension
     class Postgres
       # Returns an array of views in the database.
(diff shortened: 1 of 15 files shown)


derekprior:
This is done on top of #89 and should be merged after.

calebthompson:
Is there anywhere I can see the generated docs?

lib/scenic/adapters/postgres.rb
```diff
@@ -1,5 +1,19 @@
...
+  #
+  # Scenic ships with a Postgres adapter only but can be extended with
+  # additional adapters. The {Adapters::Postgres} adapter provides the
+  # interface.
   module Adapters
+    # An adapter for managing Postgres views.
+    #
+    # **This object is used internally by adapters and the schema dumper and is
+    # not intended to be used by application code. It is documented here for
+    # use by adapter gems.**
```
calebthompson: Will we consider this when versioning? If this API changes is that a new major version?

calebthompson:
:book:

lib/scenic/adapters/postgres.rb
derekprior: It's tagged with `@api extension` so we can set yard to not include it in generated docs by adding an exclude for it in .yardopts. We haven't done that yet. Probably should.

We can change the implementation here without worry in that case, I think. Changing the interface would depend on whether we know of any extensions.

PR #91, opened 2015-11-22
Remove errant instance variable reference

sbonami:
This commit removes an errant instance variable reference in favor of a method call in the same name.


derekprior:
Thanks for the contribution!

sbonami:
@derekprior: My pleasure!

PR #94, opened 2015-12-03
add `sql_definition` option to drop_view call

faucct:
... so reverting create_view with sql_definition would not fail

### DIFF
--- a/lib/scenic/statements.rb
+++ b/lib/scenic/statements.rb
@@ -51,7 +51,7 @@ module Scenic
     # @example Drop a view, rolling back to version 3 on rollback
     #   drop_view(:users_who_recently_logged_in, revert_to_version: 3)
     #
-    def drop_view(name, revert_to_version: nil, materialized: false)
+    def drop_view(name, revert_to_version: nil, materialized: false, sql_definition: nil)
       if materialized
         Scenic.database.drop_materialized_view(name)
       else


derekprior:
The intention is for `sql_definition` to be used only by the schema loader/dumper. In fact, we've kicked around the idea of removing it all together. I'd encourage you to use `create_view` with a specific version. The differentiator for Scenic is that it lets you keep a versioned history of your view.

If we were to accept `sql_definition` here, what would we do for `update_view`?

faucct:
I use `sql_definition` for generated sql which cannot go into files.

derekprior:
Can you give me a more concrete example? I'm not entirely following. How is the SQL generated and why couldn't it live in a file?

faucct:
In my project, I have to create 40 similar views for timeshifted program_events, so the querying would be easier. One per each utc offset. It seems easier to me to generate sql in migration file rather than generate 40 raw sql files in this case.

faucct:
So, what do you think?

derekprior:
I'm not sure just yet. There's a few tricky isses that have arisen post 1.0 and I'd like to think through those before changing any public API right now.

derekprior:
I'm inclined to say that in this case you should consider using `up` and `down` methods in your migration. I'm hesitant to make the `sql_definition` part of the API more substantial because it's really meant to exist for the schema dumper's purpose only and it's something we may decide to move away from.

If you have it handy, what is the error you got? We may need to do some work so that it raises `IrreversibleMigration` in these circumstances.

I'm going to close this for now, but please let me know if you have anything more to add to the argument.

Issue #92, opened 2015-12-03
Support concurrent materialized view refreshes


derekprior:
If you don't use `REFRESH MATERIALIZED VIEW CONCURRENTLY` then there's a chance your materialized view could be locked for selects while its refreshing. We can't just default to concurrent refreshes because in order to use `CONCURRENTLY` your materialized view has to have at least one unique index on a column. 

I dont think we should detect whether `CONCURRENTLY` can be used on a view - we should just accept an option and do as we're told.

See: http://www.postgresql.org/docs/9.4/static/sql-refreshmaterializedview.html

devonestes:
@derekprior I've addressed this issue in #97 - hopefully it's along the lines of what you were thinking.

p.s. I really love The Bike Shed - it's probably my favorite technical podcast! Thanks for putting it out every week! :radio:

derekprior:
Thanks @d

PR #97, opened 2015-12-13
Adding support to refresh materialized views concurrently

devonestes:
This PR adds support for refreshing materialized views concurrently, addressing Issue #92. I've added the necessary code to the Postgres adapter, updated the model generator to include a new `refresh_concurrently` method, and added a test to ensure that the `refresh_concurrently` method is indeed being added to new models for materialized views.

Since the ability to refresh a materialized view concurrently is dependent on a unique index without a `where` clause on that view, it is possible that calling the method will raise an exception. I thought about rescuing that exception and logging a message for the user to let them know that a concurrent refresh wasn't possible, but I felt that in this case an exception should be raised so the user doesn't think the refresh succeeded if they're not looking at their logs or for some reason doesn't see the console output. I've updated the README so there is documentation that the exception will be raised in those cases, so hopefully users won't be surprised.

There was also a small typo in `CONTRIBUTING.md`, which I've fixed here as well. It was so small I didn't think it deserved its own PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,14 +129,23 @@ those? Of course!
 
 The `scenic:view` and `scenic:model` generators accept a `--materialized`
 option for this purpose. When used with the model generator, your model will
-have the following method defined as a convenience to aid in scheduling
+have the following methods defined as conveniences to aid in scheduling
 refreshes:
 
 ```ruby
 def self.refresh
   Scenic.database.refresh_materialized_view(table_name)
 end
+
+def self.refresh_concurrently
+  Scenic.database.refresh_materialized_view_concurrently(table_name)
+end
 ```
+
+As a reminder, any materialized view that you want to refresh concurrently must
+have a unique index. If you attempt to refresh a materialized view concurrently
+without a unique index, that will raise an `ActiveRecord::StatementInvalid`
+exception.
 
 ## I don't need this view anymore. Make it go away.
 
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -77,6 +77,14 @@ module Scenic
         execute "REFRESH MATERIALIZED VIEW #{name};"
       end
 
+      # Refreshes a materialized view concurrently from its SQL schema.
+      #
+      # @param name The name of the materialized view to refresh..
+      # @return [void]
+      def refresh_materialized_view_concurrently(name)
+        execute "REFRESH MATERIALIZED VIEW CONCURRENTLY #{name};"
+      end
+
       private
 
       def execute(sql, base = ActiveRecord::Base)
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -129,15 +129,24 @@ those? Of course!
...
-have the following method defined as a convenience to aid in scheduling
+have the following methods defined as conveniences to aid in scheduling
 refreshes:
 
 ```ruby
 def self.refresh
   Scenic.database.refresh_materialized_view(table_name)
 end
+
+def self.refresh_concurrently
```
derekprior: What do you think about making this part of the existing `refresh` method? Something like:

```ruby
def self.refresh(concurently: false)
  Scenic.database.refresh_meterialzed_view(**args)
end
```

lib/scenic/adapters/postgres.rb
```diff
@@ -77,6 +77,14 @@ def refresh_materialized_view(name)
         execute "REFRESH MATERIALIZED VIEW #{name};"
       end
 
+      # Refreshes a materialized view concurrently from its SQL schema.
+      #
+      # @param name The name of the materialized view to refresh..
+      # @return [void]
+      def refresh_materialized_view_concurrently(name)
```
derekprior: If we used the earlier suggestion, this method would go away, and `refresh_materialized_view` would change its signature to `def refresh_materialized_view(name, concurently: false, **args)`.

derekprior:
Thanks so much for doing this. I had one question about the API, but I'm not 100% sure I agree with my own suggestion. I'm just a little wary of ballooning the adapter api unnecessarily, but maybe that's unfounded.

Curious what @calebthompson thinks.

README.md
devonestes: I considered that, but in the end I decided that the `if` statement that would basically switch concurrent bit on or off wasn't as neat and tidy as two separate methods. Also, since it's possible for an error to be thrown if they refresh concurrently without a unique index, I figured that the stack trace would be nicer if they saw `refresh_concurrently` instead of just something going wrong with `refresh`.

devonestes:
@derekprior Personally i could go either way, but the real thing that tipped the scale for me was having `refresh_concurrently` in the stack trace if an exception is raised. But if the consensus is that expanding the existing API to include this would be better, then I'm all for re-writing to make that happen!

derekprior:
Hmm. Excellent point. Let me think on this a bit. I don't want to introduce something just to have to reverse field.

derekprior:
@devonestes I made a few small changes and re-opened under #107 (maitaning your commit authorship).

I think using the single adapter method is a cleaner implementation and the error returned when concurrent refreshes aren't possible is descriptive enough that I'm not worried about the method name in the stack trace.

Further, we avoided having two generated methods because we felt like you were unlikely to want to sometimes refresh concurrently and other times refresh non-concurrently. The generated method for `refresh` hard codes `concurrently: false` and we'd expect users to change that where appropriate.

Thanks so much for your contribution! This will go out in 1.1.

devonestes:
@derekprior Great - thanks for the feedback and explanation! Super helpful to hear how it shook out, and I'm happy to help!

PR #102, opened 2015-12-28
Add support for updating materialized views

### LINKED ISSUE #98: Better support for updating materialized views is needed
Updating a materialized view is tricky because to change the schema you must drop and then recreate it (as far as I know). This means you lose all indexes.

For 1.0 we decided to not support updating a materialized view so that we wouldn't unwittingly drop indexes on users. We need a better long term solution for this.

I can think of a couple of solutions:

1. When `update_view` is run for a materialized view, we capture all the indexes, then re-apply them, ignoring any errors that arise due to the indexes now being invalid (those should be dropped anyway).

2. We encourage people to write their materialized view index statements as SQL in their view definition file. This might just work right now - I've just not tested it. I'm not sure if we want to go this route though. It's not obvious that it's something you'd have to do.

Any other ideas?

derekprior:
Updating materialized views used to be an error because providing a new
definition for a materialized view requires dropping and recreating it.
This will cause any indexes on the old materialized view to be dropped.
For this reason, we suggested that users manually `drop_view`,
`create_view` and `add_index`. This is a frustrating user experience.

This changes view updates to record any indexes that were in place
before the `drop_view` and then re-create those indexes once the new
materialized view is in place. Any indexes that are now invalid
(most-likely because the column no longer exists) will be dropped. This
behavior is similar to what happens when you alter the schema of a
table, so I do not believe it to be surprising.

We log the results of each index creation for the users information in
the migration output. For example:

```
== 20151228201117 UpdateSearchesToVersion2: migrating =========================
-- update_view(:searches, {:version=>2, :revert_to_version=>1, :materialized=>true})
   -> index 'searches_test_1' on 'searches' has been recreated
   -> index 'searches_test_2' on 'searches' is no longer valid and has been dropped.
   -> 0.0148s
== 20151228201117 UpdateSearchesToVersion2: migrated (0.0149s) ================
```

Closes #98

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -70,16 +70,61 @@ module Scenic
 
       # Refreshes a materialized view from its SQL schema.
       #
-      # @param name The name of the materialized view to refresh..
+      # @param name The name of the materialized view to refresh.
       # @return [void]
       def refresh_materialized_view(name)
         execute "REFRESH MATERIALIZED VIEW #{name};"
       end
 
+      # Caches indexes on the provided object before executing the block and
+      # then reapplying the indexes. Errors in applying the indexes are caught
+      # and logged to output.
+      #
+      # This is used when updating a materialized view in order to maintain all
+      # applicable indexes after the update.
+      #
+      # @param on The name of the object we are reapplying indexes on.
+      # @yield Operations to perform before reapplying indexes.
+      # @return [void]
+      def reapplying_indexes(on: name, &_block)
+        indexes = indexes_on(on)
+
+        yield
+
+        indexes.each { |index| say(try_index_create(index)) }
+      end
+
       private
 
-      def execute(sql, base = ActiveRecord::Base)
-        base.connection.execute sql
+      def connection
+        ActiveRecord::Base.connection
+      end
+
+      def execute(sql)
+        connection.execute sql
+      end
+
+      def say(message)
+        subitem = true
+        ActiveRecord::Migration.say(message, subitem)
+      end
+
+      def indexes_on(name)
+        execute(<<-SQL).map { |result| index_from_database(result) }
+          SELECT
+            t.relname as object_name,
+            i.relname as index_name,
+            pg_get_indexdef(d.indexrelid) AS definition
+          FROM pg_class t
+          INNER JOIN pg_index d ON t.oid = d.indrelid
+          INNER JOIN pg_class i ON d.indexrelid = i.oid
+          LEFT JOIN pg_namespace n ON n.oid = i.relnamespace
+          WHERE i.relkind = 'i'
+            AND d.indisprimary = 'f'
+            AND t.relname = '#{name}'
+            AND n.nspname = ANY (current_schemas(false))
+          ORDER BY i.relname
+        SQL
       end
 
       def view_from_database(result)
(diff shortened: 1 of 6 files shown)


lib/scenic/index.rb
```diff
@@ -0,0 +1,56 @@
...
+      @object_name = object_name
+      @index_name = index_name
+      @definition = definition
+    end
+
+    # Attempts to create the index, catching and logging any errors
+    #
+    # @param connection The connection to execute the SQL on
+    # @return [String] Status message for output or logging
+    def try_create(connection:)
```
derekprior: @calebthompson This was the simplest place to put this, but in the interest of modularity, should it be a separate class or a method on the adapter itself (it has postgres-specific knowledge)?

lib/scenic/index.rb
```diff
@@ -0,0 +1,56 @@
...
+      connection.execute("SAVEPOINT #{savepoint_name}")
+      connection.execute(definition)
+      connection.execute("RELEASE SAVEPOINT #{savepoint_name}")
+      "index '#{index_name}' on '#{object_name}' has been recreated"
+    rescue ActiveRecord::StatementInvalid
+      connection.execute("ROLLBACK TO SAVEPOINT #{savepoint_name}")
+      "index '#{index_name}' on '#{object_name}' is no longer valid and has been dropped."
```
derekprior: I'm fine with this.

lib/scenic/statements.rb
```diff
@@ -82,17 +80,10 @@ def update_view(name, version: nil, revert_to_version: nil, materialized: false)
         raise ArgumentError, "version is required"
       end
 
-      if materialized
-        raise ArgumentError, "Updating materialized views is not supported "\
-          "because it would cause any indexes to be dropped. Please use "\
-          "'drop_view' followed by 'create_view', being sure to also recreate "\
-          "any previously-existing indexes."
+      Scenic.database.reapplying_indexes(on: name) do
+        drop_view(name, revert_to_version: revert_to_version, materialized: materialized)
```
derekprior: I'm also fine with this. I feel like readability is better with the longer line.

lib/scenic/adapters/postgres.rb
```diff
@@ -70,16 +70,60 @@ def drop_materialized_view(name)
...
+      # Caches indexes on the provided object before executing the block and
+      # then reapplying the indexes. Errors in applying the indexes are caught
+      # and logged to output.
+      #
+      # This is used when updating a materialized view in order to maintain all
+      # applicable indexes after the update.
+      #
+      # @param on The name of the object we are reapplying indexes on.
+      # @return [void]
+      def reapplying_indexes(on: name, &block)
```
derekprior: What am I supposed to do here. I want to say this method requires a block. I also prefer `yield` to `block.call` so... should this just be ignored?

calebthompson: Well, you could use &_ or &_block. The argument doesn't actually enforce that there's a block regardless of its presence, so leaving it off seems fine to me. The `_block` is a decent compromise if you want it for documentation.

derekprior: > The argument doesn't actually enforce that there's a block regardless of its presence

Huh. I did not know that. Thanks.

lib/scenic/index.rb
calebthompson: Yeah looks like we'll want this in the adapter.

spec/smoke
```diff
@@ -97,11 +97,24 @@ main() {
...
+  rake db:migrate
+  verifySearchResults 'test'
+
   echo "rake db:rollback"
   rake db:rollback
+  rake db:rollback
   echo "[success]"
 
-  echo "rails destroy scenic:view search --materialized"
+  echo "rails destroy scenic:model search --materialized"
```
calebthompson: Looks like this line should be moved below 117, and the scenic:view echo should stay here.

derekprior: What we're actually testing here is the `model` destroy since we never tested that previously. We have to first do a `view` destroy so that the second version of the view is deleted.

calebthompson: :100:

derekprior:
@calebthompson I moved the PG specific code into the adapter. I think there's some further refactoring (as pointed out by hound's ABC warning here) to slim down and break up the adapter, but I'd like to attack that in a separate refactoring.

calebthompson:
lgtm

PR #103, opened 2015-12-29
Add FAQ section

### LINKED ISSUE #99: Document issues with `SELECT *` in views
I recently created a view that was meant to be the entirety of one table plus some aggregate data on each row. So something like this simplified, incomplete example:

```SQL
SELECT
    posts.*,
    approved_comments.count AS approved_comments_count,
    deleted_comments.count AS deleted_comments_count,
FROM
    posts
LEFT JOIN (SELECT ... ) as approved_comments ON ...
LEFT JOIN (SELECT .. ) as deleted_comments ON ...
```

Then I created a model like so:

```ruby
class PostWithStatistics < Post
end
```

This allows my `PostWithStatistics` to act as a `Post` model and lets me opt in to when I want the extra associated data. Everything works reasonably well until...

```ruby
class AddFooToPosts < ActiveRecord::Migration
  def change
    add_column :posts, :foo, :string
  end
end
```

The `posts` table now has a new column and I'd expect my `posts_with_statistics` view to reflect that because I used `SELECT *`. Unfortunately, it doesn't work this way because when the view is created with `SELECT *`, Postgres bakes into the schema the actual column names represented by `*`. To pick up the new column I'd have to use `update_view` and pass the same version number that is already deployed.

For a similar reason, this migration would also be a problem

```ruby
class RemoveBarFromPosts < ActiveRecord::Migration
  def change
    remove_column :posts, :bar, :string
  end
end
```

This migration will fail because Postgres knows that there's a view that explicitly lists the column `bar` as part of its schema. This dependency will prevent the column from being removed. The workaround is:

```ruby
class RemoveBarFromPosts < ActiveRecord::Migration
  def up
    drop_view :posts_with_statistics
    remove_column :posts, :bar, :string
    create_view :posts_with_statistics
  end
end
```

I don't think there's anything we can do from the scenic perspective to handle this. I think we likely just need to document this (and potentially link to this issue for a full description).

Anyone have any other ideas?

derekprior:
Head off some support requests with FAQs. Closes #99.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,33 @@ def change
 end
 ```
 
+## FAQs
+
+**Why is my view missing columns from the underlying table?**
+
+Did you create the view with `SELECT [table_name].*`? Most (possibly all)
+relational databases freeze the view definition at the time of creation. New
+columns will not be available in the view until the definition is updated once
+again. This can be accomplished by "updating" the view to its current definition
+to bake in the new meaning of `*`.
+
+```ruby
+add_column :posts, :title, :string
+update_view :posts_with_aggregate_data, version: 2, revert_to_version: 2
+```
+
+**When will you support MySQL?**
+
+We have no plans to add first-party support for MySQL at this time because we
+(the maintainers) do not currently have a use for it. It's our experience that
+maintaining a library effectively requires regular use of its features. We're
+not in a good position to support MySQL users.
+
+Scenic *does* support configuring different database adapters and should be
+extendable with adapter libraries. If you implement such an adapter, we're happy
+to review and link to it. We're also happy to make changes that would better
+accommodate adapter gems.
+
 ## About
 
 Scenic is maintained by [Derek Prior] and [Caleb Thompson], funded by


README.md
```diff
@@ -148,6 +148,33 @@ def change
...
+
+We have no plans to add first-party support for MySQL at this time because we
+(the maintainers) do not currently have a use for it. It's our experience that
+maintaining a library effectively requires regular use of its features. We're
+not in a good position to support MySQL users.
+
+Scenic *does* support configuring different database adapters and should be
+extendable with adapter libraries. If you implement such an adapter, we're happy
+to review and link to it. We're also happy to make changes that would better
+accommodate adapter gems.
```
calebthompson: Is the adapter api rendered as rubydoc anywhere? Could we link to that?

calebthompson:
Lgtm

README.md
derekprior: Kind of. I'm not really certain at this point how to link to rendered API docs. I am going to see if I can create a gh-pages branch with the latest release docs as part of `rake release` or something.

PR #104, opened 2015-12-29
Use better view name in examples

derekprior:
`search_results` better encapsulates what we're representing and makes a
lot more sense in ActiveRecord land: `SearchResult.where("term LIKE
%foo%")`.


calebthompson:
Lgtm

PR #107, opened 2015-12-29
Add support for concurrent refreshes

derekprior:
Concurrent refreshes allow the view to be selected from when it is being
refreshed, but require a unique index that covers all rows of the view.
For this reason, we cannot make concurrent refreshes the default.

The error returned by Postgres in situations where concurrent refreshes
are not possible is very descriptive. We test against this error both to
ensure we get a good error returned and to assert that Postgres received
the correct message.

The generated `self.refresh` model method has the `concurrently` option
hard-coded to `false` as this will work for all views and can be changed
manually where appropriate and desired. We thought about having
`refresh` take a `concurrently` option directly and passing that along,
but it seemed unlikely you'd want to sometimes refresh concurrently and
other times not.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ project codebases, issue trackers, chatrooms, and mailing lists.
 
 ## Setting Up for Development
 
-1. For the repository.
+1. Fork the repository.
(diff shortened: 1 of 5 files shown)


calebthompson:
Looks good.

PR #108, opened 2015-12-29
Add a nice landscape image to README

calebthompson:


calebthompson:
@derekprior I'm not 100% clear on what should go in a gemspec's files list, but I made it only lib/ and a few doc files.

derekprior:
I think we should leave the gemspec as is and upload the image to s3/images.thoughtbot.com. The docs render the readme, and as is, the image is broken.

calebthompson:
Woo! I got a commit in!

PR #109, opened 2015-12-29
Update docs for required postgres version

derekprior:
This also fixes build failures on travis by requiring the latest version
of Postgres there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 addons:
-  postgresql: "9.3"
+  postgresql: "9.4"
 before_install:
   - "echo '--colour' > ~/.rspec"
   - "echo 'gem: --no-document' > ~/.gemrc"
--- a/README.md
+++ b/README.md
@@ -148,8 +148,8 @@ end
 
 This will perform a non-concurrent refresh, locking the view for selects until
 the refresh is complete. You can avoid locking the view by passing
-`concurrently: true` but this requires your view to have at least one unique
-index that covers all rows.
+`concurrently: true` but this requires both PostgreSQL 9.4 and your view to have
+at least one unique index that covers all rows.
 
 ## I don't need this view anymore. Make it go away.
 
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -53,6 +53,8 @@ module Scenic
 
       # Creates a materialized view in the database
       #
+      # Materialized views require PostgreSQL 9.3 or newer.
+      #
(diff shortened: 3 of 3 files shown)


calebthompson:
Is it worth mentioning that we're not bothering to test against 9.3 anymore? This change makes it basically unsupported. Which may mean we want a Major release here instead of minor.

derekprior:
Why does it make it unsupported? You can still use all of the features you hd available yesterday. This is a new feature that will work for newer versions of postgres.

derekprior:
That is to say, there's no breaking change here.

derekprior:
> Is it worth mentioning that we're not bothering to test against 9.3 anymore?

I wouldn't mind testing against it, though in this case the differences are so small I'm unconcerned. I don't think travis supports using a matrix of postgres versions so we'd have to do something fancier. I think we just need to be cognizant of database feature support moving forward.

calebthompson:
> Why does it make it unsupported? You can still use all of the features you hd available yesterday. This is a new feature that will work for newer versions of postgres.

We don't know that, if we're not testing against 9.3 anymore.

> That is to say, there's no breaking change here.

Probably. The full API no longer works on 9.3, just a subset.

I think you're probably right and there's no need for a major change, but I can also see an argument that there ought to be.

derekprior:
> I think you're probably right and there's no need for a major change, but I can also see an argument that there ought to be.

I don't see it. 

We added new features you can opt in to at your pleasure. Depending on your particular use case you may or may not have an environment conducive to using those features, but you can continue using the exact same API you're already using with no changes, no deprecations, no breakages.

For that same reason, I didn't consider the introduction of materialized views, which required 9.3, to be a breaking change (though we were pre 1.0, so whatever).

> We don't know that, if we're not testing against 9.3 anymore.

We also don't test against Ruby 2.3, or Rails 4.1, or different patch versions of 4.2, but we support all of them. We make a tradeoff in the ease of testing and the value of the additional test coverage. Testing additional ruby versions just costs us time on travis. Testing against additional rails versions costs us time and code (our dummy app has to be runnable under the various versions of Rails). Testing against multiple Postgres versions is attractive, but I just dont see a convenient way to do it.

All this to say: I'm all for increasing our test coverage against various dependencies, but I don't think the lack of explicit tests indicates a lack of support.

derekprior:
I would also welcome a change that attempted to catch the syntax errors that `CREATE MATERIALIZED VIEW` and `REFRESH MATERIALIZED VIEW CONCURRENTLY` on versions of pg that don't support those and translates the errors into something friendly.

Issue #110, opened 2015-12-29
Raise friendly errors when features are used on versions of postgres that don't support them


derekprior:
Materialized views require PostgreSQL 9.3. Concurrent materialized view refreshes require PostgreSQL 9.4.

If those features are used on older versions of PostgreSQL, Postgres returns a syntax error. We should see if we can catch those specific messages and turn them into something a little more friendly that indicates the version of pg needs updating to use this feature.

This might be tricky or not feasible depending on the output format of the error messages.

derekprior:
Example errors from Postgres 9.2:

```
ERROR:  42601: syntax error at or near "MATERIALIZED"
LINE 1: CREATE MATERIALIZED VIEW foo AS SELECT foo'::text as text;

ERROR:  42601: syntax error at or near "REFRESH"
LINE 1: REFRESH MATERIALIZED VIEW foo
```

Example Error from Postgres 9.3:

```
ERROR:  42601: syntax error at or near "CONCURRENTLY"
LINE 1: REFRESH MATERIALIZED VIEW CONCURRENTLY tests;
```

We can do a regex match on the message. The bigger challenge is going to be testing this. We can try to fake the exception raising, but I wouldn't feel great about that. In reality, we'd need to be able to test against multiple PG versions in our suite (at least on CI). Not sure if that's possible.

devonestes:
I'd be happy to work on this, too!

Based on this closed issue on TravisCI, it doesn't look like they support that functionality: https://github.com/travis-ci/travis-ci/issues/954

Maybe one way around this would be to hook it up to several different CI servers and configure (for example) Travis to run with 9.4, CircleCI to run with 9.3 and something else like Semaphore to run with 9.2. We would probably have to have some sort of conditionals in the tests as well so that they're run or skipped depending on the Postgres version that we're testing against.

Or, if there's another CI service out there that does let you configure multiple versions of a database then you could migrate over to that service if this is a crucial part of the testing. I'd be happy to dig a little deeper and see if that's possible if you like.

derekprior:
Another idea would be to try to get the version from postgres and error based on that rather than the syntax error. I'd feel more comfortable stubbing that. The postgres adapter has a "postgresql_version" method, but unfortunately it's `protected`.

```ruby
ActiveRecord::Base.connection.send(:postgresql_version)
# => 90405
```

[...]

PR #111, opened 2015-12-30
Adding better error messages for materialized views

devonestes:
This commit adds more descriptive error messages for users when they try and use materialized views or refresh materialized views concurrently without a version of PostgreSQL that supports those operations (as mentioned in #110).

I thought about excluding checking for the possibility of an error in `drop_materalized_view` and `refresh_materalized_view` when `concurrently` is false, but for now I've decided to keep them in because they would be useful error messages for a user who was using those methods by mistake (like if they changed a particular relation from a materialized view to a view at one point and didn't change any code to refresh or drop those views).

This approach does duplicate some of the functionality that's in some protected methods in Rails, but allows us not to use those APIs and maintain compatibility with Rails 4.1.

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -1,5 +1,6 @@
 require_relative "postgres/connection"
 require_relative "postgres/views"
+require_relative "postgres/errors"
 
 module Scenic
   # Scenic database adapters.
@@ -61,6 +62,7 @@ module Scenic
       # @param sql_definition The SQL schema that defines the materialized view.
       # @return [void]
       def create_materialized_view(name, sql_definition)
+        raise_unless_materialized_views_supported
         execute "CREATE MATERIALIZED VIEW #{name} AS #{sql_definition};"
       end
 
@@ -71,6 +73,7 @@ module Scenic
       # @param name The name of the materialized view to drop.
       # @return [void]
       def drop_materialized_view(name)
+        raise_unless_materialized_views_supported
         execute "DROP MATERIALIZED VIEW #{name};"
       end
 
@@ -85,7 +88,9 @@ module Scenic
       #   error. Concurrent refreshes require PostgreSQL 9.4 or newer.
       # @return [void]
       def refresh_materialized_view(name, concurrently: false)
+        raise_unless_materialized_views_supported
         if concurrently
+          raise_unless_concurrent_refresh_supported
           execute "REFRESH MATERIALIZED VIEW CONCURRENTLY #{name};"
         else
           execute "REFRESH MATERIALIZED VIEW #{name};"
@@ -154,6 +159,18 @@ module Scenic
       rescue ActiveRecord::StatementInvalid
         connection.execute("ROLLBACK TO SAVEPOINT #{index.index_name}")
         "index '#{index.index_name}' on '#{index.object_name}' is no longer valid and has been dropped."
+      end
+
+      def raise_unless_materialized_views_supported
+        unless connection.supports_materialized_views?
+          raise MaterializedViewsNotSupportedError
+        end
+      end
+
+      def raise_unless_concurrent_refresh_supported
+        unless connection.supports_concurrent_refreshes?
+          raise ConcurrentRefreshesNotSupportedError
+        end
       end
     end
   end
--- a/spec/scenic/adapters/postgres_spec.rb
+++ b/spec/scenic/adapters/postgres_spec.rb
@@ -26,6 +26,15 @@ module Scenic
           expect(view.name).to eq("greetings")
           expect(view.materialized).to eq true
         end
+
+        it "raises an exception if the version of PostgreSQL is too old" do
+          connection = double("Connection", supports_materialized_views?: false)
+          adapter = Postgres.new(connection)
+          err = Scenic::Adapters::Postgres::MaterializedViewsNotSupportedError
+
+          expect { adapter.create_materialized_view("greetings", "select 1") }
+            .to raise_error err
+        end
       end
 
       describe "#drop_view" do
@@ -51,16 +60,46 @@ module Scenic
 
           expect(adapter.views.map(&:name)).not_to include("greetings")
         end
+
+        it "raises an exception if the version of PostgreSQL is too old" do
+          connection = double("Connection", supports_materialized_views?: false)
+          adapter = Postgres.new(connection)
+          err = Scenic::Adapters::Postgres::MaterializedViewsNotSupportedError
+
+          expect { adapter.drop_materialized_view("greetings") }
+            .to raise_error err
+        end
       end
 
       describe "#refresh_materialized_view" do
-        it "raises descriptive error if concurrent refresh is not possible" do
-          adapter = Postgres.new
-          adapter.create_materialized_view(:tests, "SELECT text 'hi' as text")
+        it "raises an exception if the version of PostgreSQL is too old" do
+          connection = double("Connection", supports_materialized_views?: false)
+          adapter = Postgres.new(connection)
+          err = Scenic::Adapters::Postgres::MaterializedViewsNotSupportedError
 
-          expect {
-            adapter.refresh_materialized_view(:tests, concurrently: true)
-          }.to raise_error(/Create a unique index with no WHERE clause/)
+          expect { adapter.refresh_materialized_view(:tests) }
+            .to raise_error err
+        end
+
+        context "refreshing concurrently" do
+          it "raises descriptive error if concurrent refresh is not possible" do
+            adapter = Postgres.new
+            adapter.create_materialized_view(:tests, "SELECT text 'hi' as text")
+
+            expect {
+              adapter.refresh_materialized_view(:tests, concurrently: true)
+            }.to raise_error(/Create a unique index with no WHERE clause/)
+          end
+
+          it "raises an exception if the version of PostgreSQL is too old" do
+            connection = double("Connection", postgresql_version: 90300)
+            adapter = Postgres.new(connection)
+            e = Scenic::Adapters::Postgres::ConcurrentRefreshesNotSupportedError
+
+            expect {
+              adapter.refresh_materialized_view(:tests, concurrently: true)
+            }.to raise_error e
+          end
         end
       end
 
--- a/lib/scenic/adapters/postgres/connection.rb
+++ b/lib/scenic/adapters/postgres/connection.rb
@@ -24,6 +24,14 @@ module Scenic
           end
         end
 
+        # True if the connection supports concurrent refreshes of materialized
+        # views.
+        #
+        # @return [Boolean]
+        def supports_concurrent_refreshes?
+          postgresql_version >= 90400
+        end
+
         # An integer representing the version of Postgres we're connected to.
         #
         # postgresql_version is public in Rails 5, but protected in earlier
--- a/lib/scenic/adapters/postgres/errors.rb
+++ b/lib/scenic/adapters/postgres/errors.rb
@@ -1,0 +1,21 @@
+module Scenic
+  module Adapters
+    class Postgres
+      # Custom error classes for PostgreSQL specific errors
+      #
+      # These error classes are for more descriptive errors when users attempt
+      # to use materialized view features not supported by their current
+      # version.
+      class MaterializedViewsNotSupportedError < StandardError
+        def initialize
+          super("Materialized views require Postgres 9.3 or newer")
+        end
+      end
+      class ConcurrentRefreshesNotSupportedError < StandardError
+        def initialize
+          super("Concurrent materialized view refreshes require Postgres 9.4 or newer")
+        end
+      end
+    end
+  end
+end
--- a/spec/scenic/adapters/postgres/connection_spec.rb
+++ b/spec/scenic/adapters/postgres/connection_spec.rb
@@ -64,6 +64,16 @@ module Scenic
           expect(connection.postgresql_version).to eq 123
         end
       end
+
+      describe "#supports_concurrent_refresh" do
+        it "is true if postgres version is at least 9.4.0" do
+          base_connection = double("Connection", postgresql_version: 90400)
+
+          connection = Postgres::Connection.new(base_connection)
+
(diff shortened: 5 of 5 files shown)


lib/scenic/adapters/postgres.rb
```diff
@@ -84,8 +86,10 @@ def drop_materialized_view(name)
       # @return [void]
       def refresh_materialized_view(name, concurrently: false)
         if concurrently
+          raise StandardError, "In order to refresh materialized views concurrently you must use PostgreSQL version 9.4 or newer" unless version_number >= 940
```
calebthompson: We also try to avoid both `unless` and "conditional surprises" at the end of lines.

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -26,6 +26,14 @@ module Adapters
           expect(view.name).to eq("greetings")
           expect(view.materialized).to eq true
         end
+
+        it "raises an exception if the version of PostgreSQL is too old" do
+          adapter = Postgres.new
+          allow(adapter).to receive(:version_number) { 920 }
```
calebthompson: You're stubbing the system under test here, which weakens the trust we can have in the test. Can we stub the `ActiveRecord::Base.connection` collaborator instead?

https://github.com/thoughtbot/scenic/pull/111/files#diff-5c0a9160f7877473c255109f580ef941R175

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -51,16 +59,44 @@ module Adapters
 
           expect(adapter.views.map(&:name)).not_to include("greetings")
         end
+
+        it "raises an exception if the version of PostgreSQL is too old" do
+          adapter = Postgres.new
+          allow(adapter).to receive(:version_number) { 920 }
```
calebthompson: https://github.com/thoughtbot/scenic/pull/111/files#r48611097

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -51,16 +59,44 @@ module Adapters
...
+            .to raise_error StandardError
+        end
       end
 
       describe "#refresh_materialized_view" do
-        it "raises descriptive error if concurrent refresh is not possible" do
+        it "raises an exception if the version of PostgreSQL is too old" do
           adapter = Postgres.new
-          adapter.create_materialized_view(:tests, "SELECT text 'hi' as text")
+          allow(adapter).to receive(:version_number) { 920 }
```
calebthompson: https://github.com/thoughtbot/scenic/pull/111/files#r48611097

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -51,16 +59,44 @@ module Adapters
...
+            adapter.create_materialized_view(:tests, "SELECT text 'hi' as text")
+
+            expect {
+              adapter.refresh_materialized_view(:tests, concurrently: true)
+            }.to raise_error(/Create a unique index with no WHERE clause/)
+          end
+
+          it "raises an exception if the version of PostgreSQL is too old" do
+            adapter = Postgres.new
+            allow(adapter).to receive(:version_number) { 930 }
```
calebthompson: https://github.com/thoughtbot/scenic/pull/111/files#r48611097

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -51,16 +59,44 @@ module Adapters
...
+            adapter.refresh_materialized_view(:tests)
+          }.to raise_error StandardError
+        end
+
+        context "refreshing concurrently" do
+          it "raises descriptive error if concurrent refresh is not possible" do
+            adapter = Postgres.new
+            adapter.create_materialized_view(:tests, "SELECT text 'hi' as text")
+
+            expect {
```
calebthompson: Feel free to ignore these, as I'm pretty sure it matches what @derekprior has been doing elsewhere.

spec/scenic/adapters/postgres_spec.rb
devonestes: Great point - updated in the most recent revision!

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -51,16 +59,42 @@ module Adapters
...
+            adapter = Postgres.new
+            adapter.create_materialized_view(:tests, "SELECT text 'hi' as text")
+
+            expect { adapter.refresh_materialized_view(:tests, concurrently: true) }
+              .to raise_error(/Create a unique index with no WHERE clause/)
+          end
+
+          it "raises an exception if the version of PostgreSQL is too old" do
+            adapter = Postgres.new
+            allow(ActiveRecord::Base.connection).to receive(:select_value) { "PostgreSQL 9.3.2 on x86_64-unknown-linux-gnu, compiled by gcc (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3, 64-bit" }
```
calebthompson: Nice. You can appease hound here by using `.and_return("...")` on a new line, indented once below `allow`.

lib/scenic/adapters/postgres.rb
```diff
@@ -166,6 +171,16 @@ def try_index_create(index)
...
         "index '#{index.index_name}' on '#{index.object_name}' is no longer valid and has been dropped."
       end
+
+      def version_number
+        @version_number ||= connection.select_value("select version()")
+          .match(/PostgreSQL (\d.\d.\d)/).captures[0].tr(".", "").to_i
+      end
+
+      def raise_unless_materialized_views_supported
+        msg = "In order to use materialized views you must use PostgreSQL version 9.3 or newer"
```
calebthompson: These are fine, of course.

lib/scenic/adapters/postgres.rb
```diff
@@ -166,6 +171,16 @@ def try_index_create(index)
...
+        raise StandardError, msg unless version_number >= 930
```
calebthompson: I think we'll still want to wrap this as `if version_number < 930`.

@derekprior do we want custom exceptions? These probably aren't being rescued so maybe not an issue?

derekprior: I think custom exceptions would clean up the code some. We wouldn't need the `msg` strings, for one. We could `raise MaterializedViewsNotSupportedError` or `raise ConcurrentRefreshesNotSupportedError`. These should be namespaced to `Scenic::Postgres`, I think.

lib/scenic/adapters/postgres.rb
```diff
@@ -166,6 +171,16 @@ def try_index_create(index)
         connection.execute("ROLLBACK TO SAVEPOINT #{index.index_name}")
         "index '#{index.index_name}' on '#{index.object_name}' is no longer valid and has been dropped."
       end
+
+      def version_number
```
derekprior: My change to make `postgresql_version` public in Rails 5 was merged. I'm tempted to make this `connection.send(:postgresql_version)`. Or maybe even check if the method exists as public, call that if it does, otherwise use send.

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -51,16 +59,42 @@ module Adapters
...
+          end
+
+          it "raises an exception if the version of PostgreSQL is too old" do
+            adapter = Postgres.new
+            allow(adapter).to receive(:version_number) { 930 }
 
-          expect {
-            adapter.refresh_materialized_view(:tests, concurrently: true)
-          }.to raise_error(/Create a unique index with no WHERE clause/)
+            expect {
```
derekprior: Yeah, I think I'll add an exception for this rule to our hound config. You can ignore this.

When I use multiline blocks that I then call methods on, it's because I had to. I don't like it any more than hound does, but I don't need the reminder.

lib/scenic/adapters/postgres.rb
devonestes: Great news! Personally I don't think the conditional to use the method if it exists is worth it (I think it might be confusing...), so I'm going to just go with `connection.send(:postgresql_version)` for now.

lib/scenic/adapters/postgres.rb
devonestes: Sounds good!

derekprior:
@devonestes Please have a gander at #115 where I had a slightly different implementation for the capability checks which I needed for a different bug. The adapter was becoming a bit of a junk drawer and that PR paves the way toward extracting classes where necessary (the index code that's in there being my next target).

derekprior:
I definitely think that's the decoration I want to go in. Would love to hear your thoughts there. Could base this work off of that PR once I get that merged (soon, I hope).

devonestes:
@derekprior That's a really good refactor! I agree, it would be much better to implement this on top #115, so once that's merge in I'll rebase and re-implement these errors.

derekprior:
115 is merged. Ready whenever you want to take a crack at it.

devonestes:
Ok, I've re-implemented the better errors after rebasing 115.

lib/scenic/adapters/postgres.rb
```diff
@@ -7,6 +7,17 @@ module Scenic
   # Scenic ships with a Postgres adapter only but can be extended with
   # additional adapters. The {Adapters::Postgres} adapter provides the
   # interface.
+  module Postgres
```
derekprior: Can we follow the same pattern I had followed with the connection and views objects? That is, extract these to `lib/scenic/adapters/postgres/errors.rb`, and require that file from here?

lib/scenic/adapters/postgres.rb
```diff
@@ -7,6 +7,17 @@ module Scenic
...
+    # Custom error classes for PostgreSQL specific errors
+    #
+    # These error classes are for more descriptive errors when users attempt
+    # to use materialized view features not supported by their current version.
+    class MaterializedViewsNotSupportedError < StandardError
```
derekprior: What do you think about having both of these supply a default message? They could do so either with:

```ruby
def initialize
  super("Here is the default message")
end
```

or:

```ruby
def message
  "here is the default message"
end
```

I think it'd be nice to say "Materialized views require Postgres 9.3 or newer" or "Concurrent materialized view refreshes require Postgres 9.4 or newer" to give some more specifics.

lib/scenic/adapters/postgres.rb
```diff
@@ -86,8 +99,10 @@ def drop_materialized_view(name)
       # @return [void]
       def refresh_materialized_view(name, concurrently: false)
         if concurrently
+          raise Scenic::Postgres::ConcurrentRefreshesNotSupportedError unless connection.supports_concurrent_refresh?
```
derekprior: We're only using this once, but I think this should be extracted to a helper like the other method just for consistency.

lib/scenic/adapters/postgres.rb
```diff
@@ -155,6 +170,10 @@ def try_index_create(index)
         connection.execute("ROLLBACK TO SAVEPOINT #{index.index_name}")
         "index '#{index.index_name}' on '#{index.object_name}' is no longer valid and has been dropped."
       end
+
+      def raise_unless_materialized_views_supported
+        raise Scenic::Postgres::MaterializedViewsNotSupportedError unless connection.supports_materialized_views?
```
derekprior: Our styleguide prefers:

```ruby
unless connection.supports_materialized_views?
  raise #..
end.
```

This makes the conditional more obvious than when it's hiding out at the end of the line - particularly a really long one like we have here.

derekprior: Additionally, if you move the files into the `Postgres` class like I had suggested above, this can just be `raise MaterializedViewsNotSupportedError`

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -26,6 +26,15 @@ module Adapters
           expect(view.name).to eq("greetings")
           expect(view.materialized).to eq true
         end
+
+        it "raises an exception if the version of PostgreSQL is too old" do
+          adapter = Postgres.new
+          allow(ActiveRecord::Base.connection)
```
derekprior: The adapter now allows us to pass in a connection, so let's take advantage of that. Something like:

```ruby
connection = double("Connection", supports_materialized_views?: false)
adapter = Postgres.new(connection)
```

I'd prefer to stub the actual method we're calling here rather than test the integration between the connection and the decorator by passing a version. We can unit test the version checks in the connection decorator specs.

spec/scenic/adapters/postgres_spec.rb
```diff
@@ -51,16 +60,46 @@ module Adapters
...
+            adapter = Postgres.new
+            adapter.create_materialized_view(:tests, "SELECT text 'hi' as text")
+
+            expect { adapter.refresh_materialized_view(:tests, concurrently: true) }
+              .to raise_error(/Create a unique index with no WHERE clause/)
+          end
+
+          it "raises an exception if the version of PostgreSQL is too old" do
+            adapter = Postgres.new
+            allow(ActiveRecord::Base.connection)
```
derekprior: similar feedback to the above here - let's pass in a double instead, and have it respond to the correct messages the way we want. Then let's unit test `supports_concurrent_refreshes` in the connection decorator specs.

lib/scenic/adapters/postgres.rb
```diff
@@ -86,8 +99,10 @@ def drop_materialized_view(name)
       # @return [void]
       def refresh_materialized_view(name, concurrently: false)
         if concurrently
+          raise Scenic::Postgres::ConcurrentRefreshesNotSupportedError unless connection.supports_concurrent_refresh?
           execute "REFRESH MATERIALIZED VIEW CONCURRENTLY #{name};"
         else
+          raise_unless_materialized_views_supported
```
derekprior: Should this be above the conditional for `concurrently`? Seems like it applies to both branches. Granted, I can't imagine ever being in a state where this error comes up, so we're just splitting hairs here...

derekprior:
This looks good. I had some style things I'd like to see cleaned up and a couple of suggestions for testing but this is close.

derekprior:
This looks great. Thanks @devonestes

derekprior:
Merged in e9b4bfb10aea633c49ed194d4b6119cade3822d6. :fireworks:

PR #115, opened 2015-12-31
Support versions of Postgres older than 9.3

### LINKED ISSUE #112: Support Postgres versions older than 9.3
Right now, our query that returns the list of views to the schema dumper assumes that `pg_matviews` exists. It does not exist on versions of Postgres < 9.3. We should be able to use the support work happening in #110 to conditionally include materialized views.

derekprior:
Postgres pre 9.3 does not support materialized views. Scenic can work
just fine for users of those versions, but it does not because the
`views` method on the adapter assumes that we can query `pg_matviews`
which does not exist on those older versions.

In this change, we introduce the concept of a decorated connection, with
the undecorated connection being injected into the adapter. This gives
us a place to decorate the connection with methods about its
capabilities. Some of these methods are defined in newer versions of
rails or are public in newer versions while being private in older
versions. We try to use the Rails versions of these methods where
possible before using our own implementation.

As part of this change, the code that looks up the views and returns
`Scenic::View` objects was extracted to its own class. This gave it some
space to grow and made for a more readable implemnentation than
otherwise possible.

Closes #112

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -1,3 +1,6 @@
+require_relative "postgres/connection"
+require_relative "postgres/views"
+
 module Scenic
   # Scenic database adapters.
   #
@@ -15,6 +18,14 @@ module Scenic
     #
     # @api extension
     class Postgres
+      # Creates an instance of the Scenic Postgres adapter
+      #
+      # @param connection The database connection the adapter should use. This
+      #   defaults to `ActiveRecord::Base.connection`
+      def initialize(connection = ActiveRecord::Base.connection)
+        @connection = Connection.new(connection)
+      end
+
       # Returns an array of views in the database.
       #
       # This collection of views is used by the [Scenic::SchemaDumper] to
@@ -22,16 +33,7 @@ module Scenic
       #
       # @return [Array<Scenic::View>]
       def views
-        execute(<<-SQL).map { |result| view_from_database(result) }
-          SELECT viewname, definition, FALSE AS materialized
-          FROM pg_views
-          WHERE schemaname = ANY (current_schemas(false))
-          AND viewname NOT IN (SELECT extname FROM pg_extension)
-          UNION
-          SELECT matviewname AS viewname, definition, TRUE AS materialized
-          FROM pg_matviews
-          WHERE schemaname = ANY (current_schemas(false))
-        SQL
+        Views.new(connection).all
       end
 
       # Creates a view in the database.
@@ -110,13 +112,8 @@ module Scenic
 
       private
 
-      def connection
-        ActiveRecord::Base.connection
-      end
-
-      def execute(sql)
-        connection.execute sql
-      end
+      attr_reader :connection
+      delegate :execute, to: :connection
 
       def say(message)
         subitem = true
@@ -139,14 +136,6 @@ module Scenic
             AND n.nspname = ANY (current_schemas(false))
           ORDER BY i.relname
         SQL
-      end
-
-      def view_from_database(result)
-        Scenic::View.new(
-          name: result["viewname"],
-          definition: result["definition"].strip,
-          materialized: result["materialized"].in?(["t", true]),
-        )
       end
 
       def index_from_database(result)
--- a/lib/scenic/adapters/postgres/connection.rb
+++ b/lib/scenic/adapters/postgres/connection.rb
@@ -1,0 +1,49 @@
+module Scenic
+  module Adapters
+    class Postgres
+      # Decorates an ActiveRecord connection with methods that help determine
+      # the connections capabilities.
+      #
+      # Every attempt is made to use the versions of these methods defined by
+      # Rails where they are available and public before falling back to our own
+      # implementations for older Rails versions.
+      #
+      # @api private
+      class Connection < SimpleDelegator
+        # True if the connection supports materialized views.
+        #
+        # Delegates to the method of the same name if it is already defined on
+        # the connection. This is the case for Rails 4.2 or higher.
+        #
+        # @return [Boolean]
+        def supports_materialized_views?
+          if undecorated_connection.respond_to?(:supports_materialized_views?)
+            super
+          else
+            postgresql_version >= 90300
+          end
+        end
+
+        # An integer representing the version of Postgres we're connected to.
+        #
+        # postgresql_version is public in Rails 5, but protected in earlier
+        # versions.
+        #
+        # @return [Integer]
+        def postgresql_version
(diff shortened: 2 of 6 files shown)


lib/scenic/adapters/postgres.rb
```diff
@@ -15,6 +18,14 @@ module Adapters
     #
     # @api extension
     class Postgres
+      # Creates an instance of the Scenic Postgres adapter
+      #
+      # @param connection The database connection the adapter should use. This
+      #   defaults to `ActiveRecord::Base.connection`
+      def initialize(connection = ActiveRecord::Base.connection)
```
bernerdschaefer: I'm not sure where / how this class is used, but does it matter that an AR connection is grabbed on initialize, versus sometime later when executing a query?

lib/scenic/adapters/postgres/connection.rb
```diff
@@ -0,0 +1,45 @@
...
+          else
+            postgresql_version >= 90300
+          end
+        end
+
+        private
+
+        # postgresql_version is public in Rails 5, but protected in earlier
+        # versions.
+        def postgresql_version
```
bernerdschaefer: Should this be public, to match the Rails 5 API being emulated?

lib/scenic/adapters/postgres.rb
derekprior: It's used in migrations and schema dumping, so not really. We expect to have an connection established by this point.

lib/scenic/adapters/postgres/views.rb
```diff
@@ -0,0 +1,56 @@
...
+        def all
+          views + materialized_views
+        end
+
+        private
+
+        attr_reader :connection
+
+        def views
+          connection.execute(<<-SQL).map { |result| to_scenic_view(result) }
```
bernerdschaefer: I'm not a huge fan of putting something other than `strip` or `strip_heredoc` after the heredoc declaration.

What about having `#views` and `#materialized_views` return the raw database arrays, and have `#all` convert them with something like `to_scenic_views(views + materialized_views)`?

spec/scenic/adapters/postgres/connection_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "spec_helper"
+
+module Scenic
+  module Adapters
+    describe Postgres::Connection do
+      describe "supports_materialized_views?" do
+        context "supports_materialized_views? was defined on connection" do
+          it "uses the previously defined version via super" do
```
bernerdschaefer: `via super` is unnecessary, and is very closely tied to the implementation.

spec/scenic/adapters/postgres/views_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      it "does not query for materialized views if not supported" do
+        connection = ActiveRecord::Base.connection
+        allow(connection).to receive(:supports_materialized_views?)
+          .and_return(false)
+        ActiveRecord::Base.connection.execute(
+          "CREATE VIEW greetings AS SELECT text 'hi' AS greeting",
+        )
+        ActiveRecord::Base.connection.execute(
+          "CREATE MATERIALIZED VIEW farewells AS SELECT text 'bye' AS text",
+        )
```
bernerdschaefer: I think means that the tests can't be run in environments where materialized views aren't supported, which would be unfortunate given the purpose of the PR. I admit I don't have a great solution to this, though.

bernerdschaefer:
I love the connection decorator, and the extraction of views into a query object.

:+1:

lib/scenic/adapters/postgres/connection.rb
derekprior: Yes, probably. Thanks!

lib/scenic/adapters/postgres/views.rb
derekprior: I'm not crazy about it either. I'll play with it and see what I can come up with.

spec/scenic/adapters/postgres/connection_spec.rb
derekprior: Agree. Will fix.

spec/scenic/adapters/postgres/views_spec.rb
derekprior: Yeah. We've been chatting about how we would even test reliably against multiple versions of postgres (no CI services support this very well). We'll cross that bridge at some point.

lib/scenic/adapters/postgres/views.rb
calebthompson: I'm in love with this syntax, for whatever I need to do after the heredoc. $0.02.

lib/scenic/adapters/postgres/views.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+
+        # All of the views that this connection has defined.
+        #
+        # This will include materialized views if those are supported by the
+        # connection.
+        #
+        # @return [Array<Scenic::View>]
+        def all
+          views + materialized_views
```
calebthompson: Can a nonmaterialized view depend on a materialized view? Is this going to get us into trouble like we saw previously with views depending on views that were alphabetically earlier than their dependancies?

derekprior: > Can a nonmaterialized view depend on a materialized view?
Yes.

> Is this going to get us into trouble like we saw previously with views depending on views that were alphabetically earlier than their dependancies?

Someday. This combination here produces no difference from the ordering that was previously in place which was a SQL union of the two queries.

But I take your point and hadn't thought of that. I pretty much figured there was no way we had all this dependency stuff sorted just yet, but I hadn't thought of exactly how we got it wrong. This is one way, for sure.

There're probably a way we can get views and matviews in a single query in the order postgres wants to return them to us (which is, hopefully, the order we'd dump them in, but that's probably not guaranteed either). But I dont think that's a change for here.

lib/scenic/adapters/postgres/views.rb
derekprior: It's the "doing something after the heredoc" that I don't like, really. It's easy to miss.

spec/scenic/adapters/postgres/connection_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+          end
+
+          connection = Postgres::Connection.new(base_connection.new)
+
+          expect(connection.postgresql_version).to eq 123
+        end
+
+        it "uses the protected method if the underlying method is not public" do
+          base_connection = Class.new do
+            protected
```
derekprior: right. It is.

PR #119, opened 2016-01-05
Refactor view updates

derekprior:
This introduces `update_view` and `update_materialized_view` as methods
on the Postgres adapter. The schema statements previously expressed
these methods in terms of `drop_view` and `create_view`, wrapped in a
call to our index reapplication code. Breaking these methods out allows
other adapters to implement these as they see fit, and also allows us to
make that index reapplication index a private implementation detail of
the adapter.

I also extracted an indexes query object (like the views query object)
and the index reapplication object to clean up the private methods on
the adapter itself.

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -1,6 +1,8 @@
 require_relative "postgres/connection"
-require_relative "postgres/views"
 require_relative "postgres/errors"
+require_relative "postgres/index_reapplication"
+require_relative "postgres/indexes"
+require_relative "postgres/views"
 
 module Scenic
   # Scenic database adapters.
@@ -40,15 +42,36 @@ module Scenic
       # Creates a view in the database.
       #
       # @param name The name of the view to create
-      # @param sql_definition the SQL schema for the view.
+      # @param sql_definition The SQL schema for the view.
+      #
       # @return [void]
       def create_view(name, sql_definition)
         execute "CREATE VIEW #{name} AS #{sql_definition};"
       end
 
+      # Updates a view in the database.
+      #
+      # This results in a {#drop_view} followed by a {#create_view}. The
+      # explicitness of that two step process is preferred to `CREATE OR
+      # REPLACE VIEW` because the former ensures that the view you are trying to
+      # update did, in fact, already exist. Additionally, `CREATE OR REPLACE
+      # VIEW` is allowed only to add new columns to the end of an existing
+      # view schema. Existing columns cannot be re-ordered, removed, or have
+      # their types changed. Drop and create overcomes this limitation as well.
+      #
+      # @param name The name of the view to update
+      # @param sql_definition The SQL schema for the updated view.
+      #
+      # @return [void]
+      def update_view(name, sql_definition)
+        drop_view(name)
+        create_view(name, sql_definition)
+      end
+
       # Drops the named view from the database
       #
       # @param name The name of the view to drop
+      #
       # @return [void]
       def drop_view(name)
         execute "DROP VIEW #{name};"
@@ -58,8 +81,9 @@ module Scenic
       #
       # @param name The name of the materialized view to create
       # @param sql_definition The SQL schema that defines the materialized view.
-      # @raise [MaterializedViewsNotSupportedError] if the version of Postgres in
-      #   use does not support materialized views.
+      #
+      # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
+      #   in use does not support materialized views.
       #
       # @return [void]
       def create_materialized_view(name, sql_definition)
@@ -67,13 +91,35 @@ module Scenic
         execute "CREATE MATERIALIZED VIEW #{name} AS #{sql_definition};"
       end
 
+      # Updates a materialized view in the database.
+      #
+      # Drops and recreates the materialized view. Attempts to maintain all
+      # previously existing and still applicable indexes on the materialized
+      # view after the view is recreated.
+      #
+      # @param name The name of the view to update
+      # @param sql_definition The SQL schema for the updated view.
+      #
+      # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
+      #   in use does not support materialized views.
+      #
+      # @return [void]
+      def update_materialized_view(name, sql_definition)
+        raise_unless_materialized_views_supported
+
+        IndexReapplication.new(connection: connection).on(name) do
+          drop_materialized_view(name)
+          create_materialized_view(name, sql_definition)
+        end
+      end
+
       # Drops a materialized view in the database
       #
       # Materialized views require PostgreSQL 9.3 or newer.
       #
       # @param name The name of the materialized view to drop.
-      # @raise [MaterializedViewsNotSupportedError] if the version of Postgres in
-      #   use does not support materialized views.
+      # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
+      #   in use does not support materialized views.
       #
       # @return [void]
       def drop_materialized_view(name)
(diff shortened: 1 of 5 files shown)


lib/scenic/adapters/postgres/index_reapplication.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        def try_index_create(index)
+          success = with_savepoint(index.index_name) do
+            connection.execute(index.definition)
+          end
+
+          if success
+            say "index '#{index.index_name}' on '#{index.object_name}' has been recreated"
+          else
+            say "index '#{index.index_name}' on '#{index.object_name}' is no longer valid and has been dropped."
```
derekprior: I am okay with this.

lib/scenic/adapters/postgres/index_reapplication.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        attr_reader :connection, :speaker
+
+        def try_index_create(index)
+          success = with_savepoint(index.index_name) do
+            connection.execute(index.definition)
+          end
+
+          if success
+            say "index '#{index.index_name}' on '#{index.object_name}' has been recreated"
```
derekprior: I'm okay with this.

lib/scenic/adapters/postgres/index_reapplication.rb
```diff
@@ -0,0 +1,63 @@
...
+        # and logged to output.
+        #
+        # @param on The name of the object we are reapplying indexes on.
+        # @yield Operations to perform before reapplying indexes. @return [void]
+        def on(name)
+          indexes = Indexes.new(connection: connection).on(name)
+
+          yield
+
+          indexes.each(&method(:try_index_create))
```
derekprior: *really* trying to make `&method` a thing... I wish ruby would happily let me drop some of those parens, but it won't. I still like it, I think.

calebthompson: I'm for it. The parens seem fine.

lib/scenic/adapters/postgres.rb
```diff
@@ -46,6 +48,16 @@ def create_view(name, sql_definition)
         execute "CREATE VIEW #{name} AS #{sql_definition};"
       end
 
+      # Updates a view in the database.
```
calebthompson: Since it's the PG adapter, I think it could be interesting to explain the internals a bit to help spread some knowledge you and I have but others might not. Proposal:

> Postgres does not allow columns to be removed or reordered from a view in an `CREATE OR REPLACE VIEW` statement. Because we'd like to allow this in update_view, we drop the view and re-create it using the new definition. See http://www.postgresql.org/docs/9.4/static/sql-createview.html for more.

lib/scenic/adapters/postgres.rb
```diff
@@ -67,6 +79,23 @@ def create_materialized_view(name, sql_definition)
         execute "CREATE MATERIALIZED VIEW #{name} AS #{sql_definition};"
       end
 
+      # Updates a materialized view in the database.
+      #
+      # Attempts to maintain all applicable indexes on the materialized view
+      # after the view is recreated.
```
calebthompson: I'd mention the raise:

```
@raise [MaterializedViewsNotSupportedError] if Postgres version is less than 9.3, which introduced materialized views.
```

Incidentally, it looks like if we really wanted to we could support this in older versions with triggers. https://wiki.postgresql.org/wiki/Materialized_Views

lib/scenic/adapters/postgres.rb
```diff
@@ -67,6 +79,23 @@ def create_materialized_view(name, sql_definition)
...
+      #
+      # @param name The name of the view to update
+      # @param sql_definition the SQL schema for the updated view.
```
calebthompson: Capitalize the definite article?

lib/scenic/adapters/postgres/index_reapplication.rb
```diff
@@ -0,0 +1,63 @@
+module Scenic
+  module Adapters
+    class Postgres
+      # Dropping a materialized view causes all indexes on the materialized view
+      # to be dropped. This object can be used to capture the existing
+      # materialized views before updating the materialized view and then
```
calebthompson: "the existing indices"

lib/scenic/adapters/postgres/index_reapplication.rb
```diff
@@ -0,0 +1,63 @@
...
+          @connection = connection
+          @speaker = speaker
+        end
+
+        # Caches indexes on the provided object before executing the block and
+        # then reapplying the indexes. Errors in applying the indexes are caught
+        # and logged to output.
+        #
+        # @param on The name of the object we are reapplying indexes on.
+        # @yield Operations to perform before reapplying indexes. @return [void]
```
calebthompson: newline before `@return`

lib/scenic/adapters/postgres/index_reapplication.rb
```diff
@@ -0,0 +1,63 @@
...
+      # @api private
+      class IndexReapplication
+        def initialize(connection:, speaker: ActiveRecord::Migration)
+          @connection = connection
+          @speaker = speaker
+        end
+
+        # Caches indexes on the provided object before executing the block and
+        # then reapplying the indexes. Errors in applying the indexes are caught
+        # and logged to output.
```
calebthompson: No they're not.

"Errors reapplying the indices cause the index not to be re-created. Each recreated or skipped index will be noted to STDOUT"

lib/scenic/adapters/postgres/indexes.rb
```diff
@@ -0,0 +1,52 @@
...
+        def on(name)
+          indexes_on(name).map(&method(:index_from_database))
+        end
+
+        private
+
+        attr_reader :connection
+
+        def indexes_on(name)
+          connection.execute(<<-SQL)
```
calebthompson: If only there was a tool that made it easy to store and execute arbitrary complex queries on a database.

lib/scenic/adapters/postgres/indexes.rb
```diff
@@ -0,0 +1,52 @@
...
+              t.relname as object_name,
+              i.relname as index_name,
+              pg_get_indexdef(d.indexrelid) AS definition
+            FROM pg_class t
+            INNER JOIN pg_index d ON t.oid = d.indrelid
+            INNER JOIN pg_class i ON d.indexrelid = i.oid
+            LEFT JOIN pg_namespace n ON n.oid = i.relnamespace
+            WHERE i.relkind = 'i'
+              AND d.indisprimary = 'f'
+              AND t.relname = '#{name}'
```
calebthompson: `name` is passed all the way down from a migration, meaning you have a (probably benign) SQL injection vulnerability here.

https://stackoverflow.com/questions/4483049/how-to-execute-a-raw-update-sql-with-dynamic-binding-in-rails has a couple of suggestions ranging from prepared statements to using [`connection.quote_table_name`](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Quoting.html#method-i-quote_table_name)

lib/scenic/adapters/postgres.rb
derekprior: Postgres does allow you to change the schema in `CREATE OR REPLACE` (it's just a drop and re-add under the covers). We opted against it because we wanted your view to have to exist first since you specifically said "update_view". Otherwise, typoing a view name for update might result in a successful migration that doesn't do what you think it was supposed to do.

derekprior: Point taken though. I will add that.

lib/scenic/adapters/postgres.rb
derekprior: Patches welcome :)

lib/scenic/adapters/postgres.rb
calebthompson: Not according to the docs

> The new query must generate the same columns that were generated by the existing view query (that is, the same column names in the same order and with the same data types), but it may add additional columns to the end of the list.

lib/scenic/adapters/postgres/indexes.rb
derekprior: I'm not too worried about injection given that the use case (migrations) but Rails itself doe use `quote_table_name` so we probably should as well. I'll leave that to another change.

derekprior:
Thanks for the great feedback. I cannot abide by `indicies`. Both are correct and PG's own documentation uses `indexes`.

lib/scenic/adapters/postgres.rb
derekprior: Huh. That's *interesting*.

derekprior: Noted BOTH reasonings

PR #121, opened 2016-01-08
Properly quote view names in SQL statements

derekprior:
This uses the existing `quote_table_name` behavior on the Postgres
connection which also handles for us the possibility that users may
include schema names.

The Postgres adapter in Rails does the same for its own schema
statements.

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -46,7 +46,7 @@ module Scenic
       #
       # @return [void]
       def create_view(name, sql_definition)
-        execute "CREATE VIEW #{name} AS #{sql_definition};"
+        execute "CREATE VIEW #{quote_table_name(name)} AS #{sql_definition};"
       end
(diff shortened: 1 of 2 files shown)


lib/scenic/adapters/postgres.rb
```diff
@@ -88,7 +88,7 @@ def drop_view(name)
       # @return [void]
       def create_materialized_view(name, sql_definition)
         raise_unless_materialized_views_supported
-        execute "CREATE MATERIALIZED VIEW #{name} AS #{sql_definition};"
+        execute "CREATE MATERIALIZED VIEW #{quote_table_name(name)} AS #{sql_definition};"
```
derekprior: My general position is that for strings like this, I'd rather have them be a shade long than wrap string creation or use a heredoc with `squish` or something. If you disagree, I'm okay with changing it.

calebthompson:
:+1:

Issue #95, opened 2015-12-06
Create views in a different database schema


ianyamey:
Often, we create views in a different database schema, so that they are logically isolated from the `public` schema and can have a different set of access rules (eg `read_only` access granted to a `reporting` user).

Is there any way to define the schema name when using scenic?

derekprior:
I believe that this should work fine, but there are a number of variables at play. Assuming the user you run migrations as has rights to the schema you want to write to, you should be able to do `create_view "my_schema.my_view"`. 

The generators don't like working with fully-qualified view names (with the ".") but everything else does work. This makes the process a little convoluted:

1. `rails generate scenic:view orders`
2. Edit the generated migration, which will say `create_view :orders` to say `create_view "my_schema.orders"`.
3. Rename the view definition file `db/views/orders_v01.sql` to `db/views/my_schema.orders_v01.sql`.

This seems to work. The process could be made smoother by updating the generators so they can accept a fully-qualified view name but are smart enough to  not try to use "CreateMySchema.Orders" as the migration class name or `MySchema.Orders` as the model name (with the model generator) but *do* use `"my_schema.orders"` as the string passed to `create_view` and `update_view`. I think this should be doable in a manner that isn't too ugly.

Give the manual process a shot and see if it will work for you. If it does and you want to take a stab at making it more generator friendly then I'd probably accept that PR so long as it can be done cleanly.

Let me know how it works out either way.

derekprior:
@ianyamey Did this work for you? If so, any interest in attempting a patch? I don't think this is a popular enough use case for me to tackle right now, but as I said - if you're willing and the patch is not too obtrusive, I'd consider merging.

PR #124, opened 2016-01-11
Adding support for views in schemas other than public

devonestes:
I ran into an issue where I wanted to create a view in a schema on Postgres other than the `public` schema, and the generator gave me files that weren't quite what I needed. It wasn't all that tricky to fix them, but I figured it would be easy enough to add in support for them for good measure.

Basically the only thing we had to do was to change the generator, and specifically we needed to override the Rails default method for `plural_file_name` so that it removed the `.` needed in the table definition when the schema was included. I also had to change the table name in the actual migration from a symbol to a string to support having a `.` in the name.

### DIFF
--- a/lib/generators/scenic/view/templates/db/migrate/create_view.erb
+++ b/lib/generators/scenic/view/templates/db/migrate/create_view.erb
@@ -1,5 +1,5 @@
 class <%= migration_class_name %> < ActiveRecord::Migration
   def change
-    create_view :<%= plural_name %><%= ", materialized: true" if materialized? %>
+    create_view <%= formatted_plural_name %><%= ", materialized: true" if materialized? %>
   end
 end
--- a/lib/generators/scenic/view/view_generator.rb
+++ b/lib/generators/scenic/view/view_generator.rb
@@ -56,7 +56,7 @@ module Scenic
 
         def migration_class_name
           if creating_new_view?
-            super
+            "Create#{class_name.gsub('.', '').pluralize}"
           else
             "Update#{class_name.pluralize}ToVersion#{version}"
           end
@@ -85,8 +85,20 @@ module Scenic
         Scenic::Definition.new(plural_file_name, previous_version)
       end
 
+      def plural_file_name
+        @plural_file_name ||= file_name.pluralize.gsub(".", "_")
+      end
+
       def destroying?
         behavior == :revoke
+      end
+
+      def formatted_plural_name
+        if plural_name.include?(".")
+          "\"#{plural_name}\""
+        else
+          ":#{plural_name}"
+        end
       end
 
       def destroying_initial_view?
--- a/lib/generators/scenic/view/templates/db/migrate/update_view.erb
+++ b/lib/generators/scenic/view/templates/db/migrate/update_view.erb
@@ -1,12 +1,12 @@
 class <%= migration_class_name %> < ActiveRecord::Migration
   def change
   <%- if materialized? -%>
-    update_view :<%= plural_name %>,
+    update_view <%= formatted_plural_name %>,
       version: <%= version %>,
       revert_to_version: <%= previous_version %>,
       materialized: true
   <%- else -%>
-    update_view :<%= plural_name %>, version: <%= version %>, revert_to_version: <%= previous_version %>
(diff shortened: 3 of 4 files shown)


derekprior:
Thanks. I don't have time to review this right now, but it looks like the PR'd solution to #95

lib/generators/scenic/view/templates/db/migrate/create_view.erb
```diff
@@ -1,5 +1,5 @@
 class <%= migration_class_name %> < ActiveRecord::Migration
   def change
-    create_view :<%= plural_name %><%= ", materialized: true" if materialized? %>
+    create_view "<%= plural_name %>"<%= ", materialized: true" if materialized? %>
```
calebthompson: It would be nice if we stuck with symbols unless we need the strings. I don't know how much I care, but it would be nice.

lib/generators/scenic/view/view_generator.rb
```diff
@@ -85,6 +85,10 @@ def previous_definition
         Scenic::Definition.new(plural_file_name, previous_version)
       end
 
+      def plural_file_name
+        @plural_file_name ||= file_name.pluralize.gsub('.', '_')
```
calebthompson: Is there a case where the table name with schema would have multiple "."s?

lib/generators/scenic/view/templates/db/migrate/create_view.erb
devonestes: We could do symbols with the quotes, like `:"non_public.searches"`, but without the quotes it ends up as `:non_public.searches`, which is a big no-no. I personally just don't like seeing symbols with quotes, but that's not for any good reason, so I'd be happy to go that route if you want.

calebthompson: Yup, my suggestion is that we keep symbols for the normal case and use strings otherwise, so a `recent_hires` view would generate `create_view :recent_hires` but a `schema.recent_hires` would generate `create_view "schema.recent_hires"`.

devonestes: Cool, I'll make that change now.

calebthompson: Since it shows up a few times, maybe add it to the view_generator as a helper that returns the quoted or symboled name?

`create_view <%= view_name %>` or whatever (I don't have a good name offhand).

derekprior:
Just wanted to mention that when I added the `quoted_table_name` stuff the other day, I came across this helper that might be of assistance here: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/utils.rb#L8

derekprior:
Sorry, that's the class it returns. I was specifically thinking about this: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/utils.rb#L8

That may or may not be helpful. Just wanted to mention it.

lib/generators/scenic/view/view_generator.rb
```diff
@@ -85,10 +85,22 @@ def previous_definition
...
+        @plural_file_name ||= file_name.pluralize.gsub(".", "_")
+      end
+
       def destroying?
         behavior == :revoke
       end
 
+      def formatted_plural_name
+        if plural_name.include?(".")
+          "\"#{plural_name}\""
```
derekprior: Let's use single quotes (outside) here to avoid escaping.

derekprior:
Looks good. Just one small comment. If hound complains about that, just ignore hound.

lib/generators/scenic/view/view_generator.rb
devonestes: If we do that then the string interpolation won't work...

```ruby
irb(main):001:0> var = 'Hi!'
=> "Hi!"
irb(main):002:0> '"#{var}"'
=> "\"\#{var}\""
```

If you don't want the escaping we could use delimiters like this: `%("#{plural_name}")` and the interpolation will still work. I don't love that syntax personally, but it's another option.

derekprior: Oh, right. That makes sense. nevermind. I'll merge this next time I'm able.

derekprior:
Merged in 6eed824. Thank you!

PR #126, opened 2016-01-13
Add last method error workaround to FAQs

dixpac:
screen shot 2016-01-13 at 20 18 41

```last``` method uses ```ORDER BY id DESC``` to find last row. When we explicitly don't set primary key in our model, there is error :(
Not sure if this FAQ should go in separate section or we could somehow stick it with ```find``` FAQ, but for me this is more readable :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,17 @@ class People < ActiveRecord::Base
   self.primary_key = :id
 end
 ```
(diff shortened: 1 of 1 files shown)


derekprior:
Thanks for bringing this to our attention. I combined it with the previous FAQ here: e632bc7f4904b885c9943c852e4cc33299f16666

Issue #129, opened 2016-01-27
ActiveRecord::StatementInvalid: PG::ConnectionBad: connection is closed when using db:schema:load


tatyree:
El Capitan 
Rails 4.2.5
Postgres 9.4.5 from homebrew

Create a new rails project
Add the scenic gem and bundle 
`bundle exec rails g scenic:view hello_world`
Add anything to the view (in this instance `SELECT 'hello world';`)
`bundle exec rake db:create db:migrate` works as expected
Subsequent calls to 
`bundle exec rake db:drop db:setup` or `bundle exec rake db:drop db:create db:schema:load` fail with:

```ruby
ActiveRecord::StatementInvalid: PG::ConnectionBad: connection is closed: CREATE VIEW "hello_worlds" AS       SELECT 'hello world';
```
However, subsequent `bundle exec rake db:drop db:create db:migrate` calls work.

derekprior:
We do precisely this in our test suite with Rails 4.2.5 and PG 9.4. I'm also using this on my own Rails 4.2.5 projects without issue. I'll give this a shot on Friday to see if I can reproduce here.

adamnbowen:
I can reproduce this issue by following the steps above with Postgres.app at version 9.4.4.0 (I'm also on El Capitan), so it at least shouldn't be Homebrew-specific.

AskarZinurov:
Reproduced on 4.1.14 & PG 9.4

derekprior:
I created a test app to reproduce this: https://github.com/derekprior/scenic_bug

And then I executed the rake tasks with `-t` so I could get a trace:

```
scenic_bug master % bundle exec rake db:reset -t
** Invoke db:reset (first_time)
** Invoke environment (first_time)
** Execute environment
** Invoke db:load_config (first_time)
** Execute db:load_config
** Execute db:reset
** Invoke db:drop (first_time)
** Invoke db:load_config
** Execute db:drop
** Invoke db:setup (first_time)
** Invoke db:schema:load_if_ruby (first_time)
** Invoke db:create (first_time)
** Invoke db:load_config
** Execute db:create
** Invoke environment
** Execute db:schema:load_if_ruby
** Invoke db:schema:load (first_time)
** Invoke environment
** Invoke db:load_config
** Execute db:schema:load
-- enable_extension("plpgsql")
   -> 0.0172s
-- create_view(:hello_worlds, {:sql_definition=>"      SELECT 'hello world';\n"})
[...]

PR #130, opened 2016-01-29
Get a fresh connection for all operations

derekprior:
The Postgres adapter was previously accepting an ActiveRecord connection
directly. In some operations, this resulted in Scenic attempting to use
a connection that had already been closed by Rails.

Instead of taking a connection directly, we now take something that
responds to a `connection`, returning a fresh connection.

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -27,15 +27,15 @@ module Scenic
       # {Scenic.configure} is not required, but the example below shows how one
       # would explicitly set it.
       #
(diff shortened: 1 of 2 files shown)


derekprior:
Closes #129

calebthompson:
Yeah, looks good. I assume you tested this against the issue in #129?

If it doesn't fix it, we can also call `establish_connection` with `Rails.configuration.database_configuration` as the argument.

derekprior:
Yeah, I tested with the app I created to reproduce that. Looks good.

calebthompson:
Ship.

PR #135, opened 2016-02-05
Fix materialized view issues

derekprior:
The individual commits have more detail (and I suggest reviewing them), but briefly:

1. We were not dumping materialized view indexes on `db:schema:dump`.
2. Our index reapplication on materialized view code was broken befcause our query to find indexes was incorrect.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -13,13 +13,14 @@ appraise "rails42" do
   gem "railties", "~> 4.2.0"
 end
 
-appraise "rails50" do
-  gem "activerecord", "~> 5.0.0.beta1"
-  gem "railties", "~> 5.0.0.beta1"
-  gem "rspec-rails", github: "rspec/rspec-rails"
-  gem "rspec-support", github: "rspec/rspec-support"
-  gem "rspec-core", github: "rspec/rspec-core"
-  gem "rspec-mocks", github: "rspec/rspec-mocks"
-  gem "rspec-expectations", github: "rspec/rspec-expectations"
-  gem "rspec", github: "rspec/rspec"
+if RUBY_VERSION > "2.2.0"
+  appraise "rails50" do
+    gem "rails", github: "rails/rails"
+    gem "rspec-rails", github: "rspec/rspec-rails"
+    gem "rspec-support", github: "rspec/rspec-support"
+    gem "rspec-core", github: "rspec/rspec-core"
+    gem "rspec-mocks", github: "rspec/rspec-mocks"
+    gem "rspec-expectations", github: "rspec/rspec-expectations"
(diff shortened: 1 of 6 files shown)


derekprior:
Argh. Rails 5 changed index dumping so it's inside the `create_table` block:
`t.index ["results"], name: "index_searches_on_results", using: :btree`

Back to the drawing board...

derekprior:
Okay, we're all set now. I have no clue why the builds under ruby 2.1.8 only are failing. They are having trouble installing the pg gem. I don't have this problem locally - I even installed 2.1.8 and ran through appraisal to be sure. I don't want to hold up these fixes whil I dive into the impenetrable world of travis. @calebthompson What do you think?

lib/scenic/view.rb
```diff
@@ -44,10 +44,10 @@ def ==(other)
     def to_schema
       materialized_option = materialized ? "materialized: true, " : ""
       <<-DEFINITION
-
   create_view :#{name}, #{materialized_option} sql_definition: <<-\SQL
     #{definition.indent(2)}
   SQL
+
       DEFINITION
```
geoffharcourt: Does `strip_heredoc` let you write this without the reverse-indentation that breaks back to the left?

geoffharcourt:
If this passes locally on Ruby 2.1.8 and all other combinations in the matrix are green, I think this is OK to ship. I had one comment about formatting that might make the code _look_ better, but nothing functional to change.

Appraisals
```diff
@@ -13,13 +13,14 @@ appraise "rails42" do
...
-appraise "rails50" do
-  gem "activerecord", "~> 5.0.0.beta1"
-  gem "railties", "~> 5.0.0.beta1"
-  gem "rspec-rails", github: "rspec/rspec-rails"
-  gem "rspec-support", github: "rspec/rspec-support"
-  gem "rspec-core", github: "rspec/rspec-core"
-  gem "rspec-mocks", github: "rspec/rspec-mocks"
-  gem "rspec-expectations", github: "rspec/rspec-expectations"
-  gem "rspec", github: "rspec/rspec"
+if RUBY_VERSION > "2.2.0"
```
geoffharcourt: I added an [idea on Appraisals](https://github.com/thoughtbot/appraisal/issues/104) to avoid having to do this in the future.

lib/scenic/view.rb
derekprior: I've played with stip_heredoc so much and this is the most reasonable looking output I've been able to get for schema.rb.

derekprior:
Looks like we're not alone with 2.1.8 issues: https://github.com/travis-ci/travis-ci/issues/5580

PR #145, opened 2016-02-18
Add update_view_replace migration statement

mordocai:
Despite the fact that drop_view, create_view is usually the better
choice there are some situations where CREATE OR REPLACE VIEW can be
better. An example is when you have a complicated dependency tree
between your views or when the view is depended upon by a slow
materialized view.

This commit adds a update_view_replace migration statement (and
associated code) that uses CREATE OR REPLACE VIEW instead of
dropping and creating the view.

I understand you had reasons for not having this functionality, so I have made this pull request early. If this functionality is just not wanted, I'd rather not waste my time making a better patch.

Assuming you are open to adding this feature, I will spend the time working on updating tests, docs, etc for the new method.

I am also open to naming it something else of course, this is just the first name that came to mind.

A possibility is adding a flag to update_view but I felt this was more clear.

Edit: We "needed" this for work and currently are using this for a feature branch.

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -81,6 +81,29 @@ module Scenic
         create_view(name, sql_definition)
       end
 
+      # Updates a view in the database using `CREATE OR REPLACE VIEW`.
+      #
+      # This results in a `CREATE OR REPLACE VIEW`. Most of the time the
+      # explicitness of the two step process used in {#update_view} is preferred to
+      # `CREATE OR REPLACE VIEW` because the former ensures that the view you are trying
+      # to update did, in fact, already exist. Additionally, `CREATE OR REPLACE
+      # VIEW` is allowed only to add new columns to the end of an existing
+      # view schema. Existing columns cannot be re-ordered, removed, or have
+      # their types changed. Drop and create overcomes this limitation as well.
+      #
+      # However, when there is a tangled dependency tree `CREATE OR REPLACE VIEW` can
+      # be preferable.
+      #
+      # This is typically called in a migration via {Statements#update_view_replace}.
+      #
+      # @param name The name of the view to update
+      # @param sql_definition The SQL schema for the updated view.
+      #
+      # @return [void]
+      def update_view_replace(name, sql_definition)
+        execute "CREATE OR REPLACE VIEW #{quote_table_name(name)} AS #{sql_definition};"
+      end
+
       # Drops the named view from the database
       #
       # This is typically called in a migration via {Statements#drop_view}.
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -81,26 +81,28 @@ module Scenic
         create_view(name, sql_definition)
       end
 
-      # Updates a view in the database using `CREATE OR REPLACE VIEW`.
+      # Replaces a view in the database using `CREATE OR REPLACE VIEW`.
       #
       # This results in a `CREATE OR REPLACE VIEW`. Most of the time the
-      # explicitness of the two step process used in {#update_view} is preferred to
-      # `CREATE OR REPLACE VIEW` because the former ensures that the view you are trying
-      # to update did, in fact, already exist. Additionally, `CREATE OR REPLACE
-      # VIEW` is allowed only to add new columns to the end of an existing
-      # view schema. Existing columns cannot be re-ordered, removed, or have
-      # their types changed. Drop and create overcomes this limitation as well.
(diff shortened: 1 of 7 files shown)


derekprior:
Seems like something we should have but I don't love the name. I'd be okay with `replace_view`, I think. I  could also be convinced to use a keyword arg, but having it's own method gives us more space to document it and its differences from `update_view`. What do you think @calebthompson

lib/scenic/statements.rb
```diff
@@ -89,6 +89,32 @@ def update_view(name, version: nil, revert_to_version: nil, materialized: false)
       end
     end
 
+    # Update a database view to a new version using `CREATE OR REPLACE VIEW`.
+    #
+    # The existing view is created/replaced using the supplied `version`
+    # parameter.
+    #
+    # Does not work with materialized views, no parameter for that is provided!
```
derekprior: What is the error I get if I call this with `materialized: true` on a materialized view? What is the error if I call it without the keyword but the view is materialized? I think it might be worth accepting that as a parameter and giving a nice, specific error if it'll help in these cases. What do you think?

lib/scenic/adapters/postgres.rb
```diff
@@ -81,6 +81,29 @@ def update_view(name, sql_definition)
         create_view(name, sql_definition)
       end
 
+      # Updates a view in the database using `CREATE OR REPLACE VIEW`.
+      #
+      # This results in a `CREATE OR REPLACE VIEW`. Most of the time the
+      # explicitness of the two step process used in {#update_view} is preferred to
+      # `CREATE OR REPLACE VIEW` because the former ensures that the view you are trying
+      # to update did, in fact, already exist. Additionally, `CREATE OR REPLACE
```
derekprior: We *could* verify this as well if we want to avoid calling this method to replace a non-existent view. What do you think?

mordocai: I think i'd go with the principle of least surprise here. If the name ends up being replace_view (which sounds fine to me) then I'd say giving an error if the view doesn't exist is probably a good thing.

lib/scenic/statements.rb
mordocai: I like the idea, I'll take a look at a good way to handle it and provide a nice error.

mordocai:
Since it sounds like this will be well received, i'll start working on cleaning it up during my innovation days at work.

lib/scenic/adapters/postgres.rb
```diff
@@ -81,6 +81,29 @@ def update_view(name, sql_definition)
...
+      # be preferable.
+      #
+      # This is typically called in a migration via {Statements#update_view_replace}.
+      #
+      # @param name The name of the view to update
+      # @param sql_definition The SQL schema for the updated view.
+      #
+      # @return [void]
+      def update_view_replace(name, sql_definition)
+        execute "CREATE OR REPLACE VIEW #{quote_table_name(name)} AS #{sql_definition};"
```
mordocai: houndci is apparently ignoring this rule in places (like the refresh materialized view esecute line above this in the file). I don't see how this is configured.

@derekprior Should this be broken into multiple lines or should we make an exception with hound or something else? If broken into multiple lines, how do you guys like to do that for your strings?

lib/scenic/statements.rb
```diff
@@ -89,6 +89,32 @@ def update_view(name, version: nil, revert_to_version: nil, materialized: false)
...
+    # @param name [String, Symbol] The name of the database view.
+    # @param version [Fixnum] The version number of the view.
+    # @param revert_to_version [Fixnum] The version number to rollback to on
+    #   `rake db rollback`
+    # @return The database response from executing the create statement.
+    #
+    # @example
+    #   update_view_replace :engagement_reports, version: 3, revert_to_version: 2
+    #
+    def update_view_replace(name, version: nil, revert_to_version: nil)
```
mordocai: This is used by Scenic::CommandRecorder#perform_scenic_inversion. Do we want to add an exception in hound? @derekprior

mordocai:
I'm writing some documentation for the readme now, and had a thought. Should this have a generator too? The rest of the docs don't have to talk about modifying the migrations themselves at all, but currently to do replace you'll have to do so.

mordocai:
This in its current state has the suggestions besides checking if a view exists. Please re-review @derekprior @calebthompson

calebthompson:
Merged in 621cba4f4556103c9efb277187d8a3bb7ec198e3

calebthompson:
Whoops wrong one

calebthompson:
Merged in 160e05db12209a1236b8a41aa08bde2a362ef34e

PR #150, opened 2016-04-07
Interpret embedded ERB in sql files

adzap:
I needed to use some logic for generating a complex view. So using the Rails-ish way of interpreting ERB in these kinds of text files seemed like a reasonable solution.

Thoughts?

### DIFF
--- a/lib/scenic/definition.rb
+++ b/lib/scenic/definition.rb
@@ -1,3 +1,5 @@
+require 'erb'
+
 module Scenic
   # @api private
   class Definition
@@ -7,7 +9,7 @@ module Scenic
     end
 
     def to_sql
-      File.read(full_path).tap do |content|
+      ERB.new(File.read(full_path)).result.tap do |content|
         if content.empty?
           raise "Define view query in #{path} before migrating."
         end
--- a/lib/scenic/definition.rb
+++ b/lib/scenic/definition.rb
@@ -1,4 +1,4 @@
-require 'erb'
+require "erb"
 
 module Scenic
   # @api private
--- a/spec/scenic/definition_spec.rb
+++ b/spec/scenic/definition_spec.rb
@@ -19,6 +19,15 @@ module Scenic
           Definition.new("searches", 1).to_sql
         end.to raise_error RuntimeError
       end
+
+      it "interprets embeded ERB statements" do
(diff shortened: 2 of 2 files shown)


derekprior:
What's the specific use case here? I'm concerned because while this might be convenient at the time you're writing the view, drift of the interpolated values may result in different users having different views. Consider this:

```ruby
class Order
  SHIPPED_STATUS = "complete".freeze
end
```

```erb
SELECT * FROM orders WHERE status = "<%= Order::SHIPPED_STATUS %>"
```

If that constant changes and one developer re-migrates they now have a different view than you had when you first wrote the view. And that new view has been dumped to structure.sql too.

I'm worried the win here isn't big enough to justify this risk. It seems similar to the reason I'd try to avoid using ActiveRecord models in my migrations, but maybe I'm missing something?

derekprior:
Also, if we do this (and I'm willing to be convinced), do we want to have it be explicit via the `.erb` file extension? If we process every SQL definition as ERB then any `<%` in the definition could cause problems. This seems an unlikely character combination to have in a view definition, but there'd be no way to work around it.

adzap:
My use case is to create a lengthy view from existing knowledge in the app, table names, column names and other metadata, then use some loops to generate the final columns which are pg functions. To avoid transcription errors I wanted to use Ruby. Initially I was going to generate it outside of scenic and dump it in the view, but it was very verbose. But I wondered why it needed to much different from anything I could do in a regular migration which contains Ruby to generate SQL behinds the scenes.

The concern about value drift in the migration SQL is valid in the context of it being able to be affected from distant code in the app. Which is a problem in any app design; a change in one location having unintended consequences in another. Though a general concern about value changes doesn't seem valid. Such changes always occur and a remigration to use the new value may be the exact intention, nevertheless, the code change would/should be source controlled. Your example of fragile information in such a migration such as model data is rather fraught and to be avoided, and again an issue with developer common sense.

In general not seeing the final SQL before migrating is a nothing new to Rails. That is how the migrations already work. Though the difference here is that no formal DSL is present to attempt represent the output SQL.

Adding the `.erb` is a good idea given your example.

calebthompson:
I'm sorry, but we're not convinced this is a good idea. The chance of the SQL being generated changing outweighs the benefits in our opinion. It might be more useful to generate the actual .sql file with ERB *once* and commit the results but not the source, but that's outside the scope of Scenic.

PR #151, opened 2016-04-29
Singularize file_path when injecting into class

### LINKED ISSUE #148: Issue with inconsistent singularization with materialization
If the generator is using singularization it forgets to do so on the materialization insert.

```ruby
rails generate scenic:model inventory_types --materialized
[WARNING] The model name 'inventory_types' was recognized as a plural, using the singular 'inventory_type' instead. Override with --force-plural or setup custom inflection rules for this noun before running the generator.
      invoke  active_record
      create    app/models/inventory_type.rb
      invoke    rspec
      create      spec/models/inventory_type_spec.rb
      invoke      factory_girl
      create        spec/factories/inventory_types.rb
      insert  app/models/inventory_types.rb
/Users/sirwolfgang/.gem/ruby/2.3.0/gems/thor-0.19.1/lib/thor/actions/inject_into_file.rb:98:in `binread': No such file or directory @ rb_sysopen - /Users/sirwolfgang/Projects/eve/app/models/inventory_types.rb (Errno::ENOENT)```

calebthompson:
Railties assumes that `file_name`, set in
[`assign_names!`](https://github.com/rails/rails/blob/4a446220a67e115c04a2b562dce92981a8e18c78/railties/lib/rails/generators/named_base.rb#L177-L181)
as a derivative of `name` which can be plural or singular, will always be
singular when it implicitly aliases it in
[`singular_name`](https://github.com/rails/rails/blob/4a446220a67e115c04a2b562dce92981a8e18c78/railties/lib/rails/generators/named_base.rb#L36-L38).

This is really a rails/rails issue, but to resolve #148 for ourselves we'll
call `singularize` on `file_path` which inherits this assumption. It seems to
me that the solution really would be to singularize `name` in NamedBase before
calling `assign_names` since every instance of
`plural_*` calls pluralize.

### DIFF
--- a/lib/generators/scenic/model/model_generator.rb
+++ b/lib/generators/scenic/model/model_generator.rb
@@ -11,12 +11,12 @@ module Scenic
       source_root File.expand_path("../templates", __FILE__)
 
       def invoke_rails_model_generator
-        invoke "model", [name], options.merge(migration: false)
+        invoke "model", [file_path.singularize], options.merge(migration: false)
       end
 
       def inject_model_methods
         if materialized? && generating?
-          inject_into_class "app/models/#{file_path}.rb", class_name do
+          inject_into_class "app/models/#{file_path.singularize}.rb", class_name do
             evaluate_template("model.erb")
           end
         end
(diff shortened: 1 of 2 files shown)


derekprior:
Looks good to me, but the tests are failing

PR #155, opened 2016-06-03
Add binstubs for `rake` and `rspec`

teoljungberg:
The binstubs are generated by:

    % bundle binstub bundle rspec-core


bin/setup
```diff
@@ -5,6 +5,9 @@ set -e
 gem install bundler --conservative
 bundle check || bundle install
 
+# Add binstubs to PATH via export PATH=".git/safe/../../bin:$PATH" in ~/.zshenv
+mkdir -p .git/safe
```
calebthompson: It doesn't make much sense to use this approach (trusting a repository) if you do it automatically. This should be a deliberate choice by each person to trust the repo after examining the code and bins.

teoljungberg: Valid concern, I took what we have in our [`bin/setup` for Suspenders] which also is taken from our [shell snippets]

[shell snippets]: https://github.com/thoughtbot/shell-snippets/blob/master/generic/add-bin-to-path.sh
[`bin/setup` for suspenders]: https://github.com/thoughtbot/suspenders/blob/b3ee72c041d017c21bfc1970da20a9d36aed0f9f/templates/bin_setup#L16-L17

calebthompson: Yup, and while I disagree it should be there as well, it's scarier in an open source project that might be using tb's dotfiles directly. In a closed source project only a few people control it's a more understandable risk to take, but in an open source project with more contributors it scares me too much.

derekprior: I concur with Caleb.

Though if you are running a bin/setup file you haven't inspected, creating
.git/safe is the least of your worries.

teoljungberg: Good point - remedied in b6a928c

calebthompson:
LGTM

teoljungberg:
Thanks, Caleb and Derek.

PR #156, opened 2016-06-07
Allow custom path to be used for migration and view files

craigsheen:
In a project I have a number of different databases setup, each database has it's own folder and I have a rake task which sets runs through the different databases and runs the migrations for them. The structure looks like below;

```
# apps main db
- db
  - schema.rb
  - migrate
    - 123_migration_1.rb
    - 456_migration_2.rb
# other database
- db_foobar
  - schema.rb
  - migrate
    - 098_migration_1.rb
    - 765_migration_2.rb
```
I have the need to create a view inside the other database and for this the migrations and the views need to go into the `db_foobar` folder. I can drag them into the correct folder once we create them, however as scenic hardcodes the definition path to `db/migrate` this fails.

This PR adds the ability to either manually add `custom_path` to the scenic migrations (`create_view`. `update_view` etc.) or pass this in as an option to the generator (`rails g scenic:view my_view --custom_path=db_foobar`)

This will not effect the default use of scenic at all.

Let me know your thoughts on if this would be of interest and if any changes are needed.

### DIFF
--- a/lib/generators/scenic/custom_pathable.rb
+++ b/lib/generators/scenic/custom_pathable.rb
@@ -8,8 +8,8 @@ module Scenic
         class_option :custom_path,
           type: :string,
           required: false,
-          desc: 'Sets the path for the view',
-          default: ''
+          desc: "Sets the path for the view",
+          default: ""
       end
 
       private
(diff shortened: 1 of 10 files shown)


calebthompson:
I think that in the absence of any other users for this change, we're going to have to suggest you monkey patch Scenic, use this fork, etc. rather than merging this, as it adds a fair bit of complexity we'd need to maintain moving forward.

PR #160, opened 2016-06-30
Don't lock materialized views when updating them

richddr:
Update to /lib/scenic/adapters/postgres.rb#update_materialized_view

**Before if there were big changes done to a materialized view, when that migration was run on a production environment, it would lock up the materialized view and it would result in end users having a bad experience. This PR addresses this issue by doing the following:**

When updating a materialized view, instead of:
- Dropping the materialized view
- Recreating the materialized view from the latest version

Do the following:
- Create a materialized view from the latest version with a temporary name ("#{materialized_view_name}_temp").
   This includes applying the new indexes contained in the new version.
- Drop the old materialized view
- Rename the newly created temporary view to the original name

I saw that there weren't any specs for `#update_materialized_view` so I did not write any, also this would be hard to test. All existing specs passed locally (`bin/appraisal rake`). You can verify that this works by having a v1 of a view, then on the v2 of it make it really complex (completely different SELECT, indexes) so it takes a good amount of time to run that migration, while the migration is running, connect to the database and query the materialized view and verify that its not locked and you can get results from it (previous version).

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,12 @@ a new version of it.
 This is not desirable when you have complicated hierarchies of views, especially
 when some of those views may be materialized and take a long time to recreate.
 
-You can use `replace_view` to generate a CREATE OR REPLACE VIEW SQL statement.
+You can use `replace_view`. `replace_view` generates a CREATE OR REPLACE VIEW
+SQL statement on a normal view. For a materialized view, it creates a new
+materialized view under a temporary name (including the index definition
+specified in that version of the materialized view), then it drops the previous
+existing materialized view and then renames the new materialized view to its
+original name.
 
 See postgresql documentation on how this works:
 http://www.postgresql.org/docs/current/static/sql-createview.html
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -136,9 +136,11 @@ module Scenic
 
       # Updates a materialized view in the database.
       #
-      # Drops and recreates the materialized view. Attempts to maintain all
-      # previously existing and still applicable indexes on the materialized
-      # view after the view is recreated.
+      # Recreate the materialized view under a temporary name and apply new
+      # indexes (if any). Drop previous materialized view, renames the
+      # temporary materialized view to its original name. Attempts to maintain
+      # all previously existing and still applicable indexes on the
+      # materialized view after the view is recreated.
       #
       # This is typically called in a migration via {Statements#update_view}.
       #
@@ -152,9 +154,22 @@ module Scenic
       def update_materialized_view(name, sql_definition)
         raise_unless_materialized_views_supported
 
+        temp_mv_name = "#{name}_temp"
         IndexReapplication.new(connection: connection).on(name) do
+          mv_definition = get_definition_for(:create_definition, sql_definition)
+          create_materialized_view(temp_mv_name, mv_definition.first)
+
+          new_index_definitions = get_definition_for(:indexes, sql_definition)
+
+          new_index_definitions.each do |idx_definition|
+            execute idx_definition.gsub(name, temp_mv_name)
+          end
+
           drop_materialized_view(name)
-          create_materialized_view(name, sql_definition)
+
+          execute "ALTER MATERIALIZED VIEW #{quote_table_name(temp_mv_name)} RENAME TO #{quote_table_name(name)};"
+
+          update_index_names_for(temp_mv_name, name)
         end
       end
 
@@ -235,6 +250,35 @@ module Scenic
           self,
           connection,
         )
+      end
+
+      def indexes_for(tbl)
+        execute "SELECT indexname FROM pg_indexes WHERE tablename = '#{tbl}'"
+      end
+
+      def update_index_names_for(old_tbl, tbl)
+        indexes_for(tbl).map{ |r| r['indexname'] }.each do |indexname|
+          new_indexname = indexname.gsub(old_tbl, tbl)
+
+          execute "ALTER INDEX #{indexname} RENAME TO #{new_indexname}" unless indexname == new_indexname
+        end
+      end
+
+      def get_definition_for(type, sql_definition)
+        command =
+          case type
+          when :create_definition then :reject
+          when :indexes then :select
+          end
+
+        sql_definition.
+          strip.
+          split(';').
+          send(command) do |sql_line|
+            sql_line.
+              gsub(/\n/,'').
+              start_with?('CREATE INDEX', 'CREATE UNIQUE INDEX')
+          end
       end
     end
   end
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -154,23 +154,48 @@ module Scenic
       def update_materialized_view(name, sql_definition)
         raise_unless_materialized_views_supported
 
-        temp_mv_name = "#{name}_temp"
         IndexReapplication.new(connection: connection).on(name) do
-          mv_definition = get_definition_for(:create_definition, sql_definition)
-          create_materialized_view(temp_mv_name, mv_definition.first)
-
-          new_index_definitions = get_definition_for(:indexes, sql_definition)
-
-          new_index_definitions.each do |idx_definition|
-            execute idx_definition.gsub(name, temp_mv_name)
-          end
-
-          drop_materialized_view(name)
-
-          execute "ALTER MATERIALIZED VIEW #{quote_table_name(temp_mv_name)} RENAME TO #{quote_table_name(name)};"
-
-          update_index_names_for(temp_mv_name, name)
+          replace_materialized_view(name, sql_definition)
         end
+      end
+
+      # Replaces a materialized view in the database.
+      #
+      # Recreate the materialized view under a temporary name and apply new
+      # indexes (if any). Drop previous materialized view, renames the
+      # temporary materialized view to its original name.
+      #
+      # This is typically called in a migration via {Statements#replace_view}.
+      #
+      # @param name The name of the view to update
+      # @param sql_definition The SQL schema for the updated view.
+      #
+      # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
+      #   in use does not support materialized views.
+      #
+      # @return [void]
+      def replace_materialized_view(name, sql_definition)
+        raise_unless_materialized_views_supported
+
+        temp_mv_name = "#{name}_temp"
+
+        mv_definition = get_definition_for(:create_definition, sql_definition)
+        create_materialized_view(temp_mv_name, mv_definition.first)
+
+        new_index_definitions = get_definition_for(:indexes, sql_definition)
+
+        new_index_definitions.each do |idx_definition|
+          execute idx_definition.gsub(name, temp_mv_name)
+        end
+
+        drop_materialized_view(name)
+
+        sql = "ALTER MATERIALIZED VIEW #{quote_table_name(temp_mv_name)} " +
+          "RENAME TO #{quote_table_name(name)};"
+
+        execute sql
+
+        update_index_names_for(temp_mv_name, name)
       end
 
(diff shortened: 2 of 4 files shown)


timz1290:
It'd be nice to use the same logic here (creating a temp materialized view, drop the original, rename the temp to original) to implement replace_view, materialized:true

richddr:
That sounds like a good idea, will attempt to implement in this PR.

richddr:
Can't see the CodeClimate issues.

lib/scenic/adapters/postgres.rb
```diff
@@ -226,6 +264,35 @@ def raise_unless_concurrent_refresh_supported
...
+
+      def indexes_for(tbl)
+        execute "SELECT indexname FROM pg_indexes WHERE tablename = '#{tbl}'"
+      end
+
+      def update_index_names_for(old_tbl, tbl)
+        indexes_for(tbl).map { |r| r["indexname"] }.each do |indexname|
+          new_indexname = indexname.gsub(old_tbl, tbl)
+
+          execute "ALTER INDEX #{indexname} RENAME TO #{new_indexname}" unless indexname == new_indexname
```
calebthompson: Please don't use 'conditional surprises', preferring the block format of conditions. Also please invert the conditional to `if index_name != new_index_name`

README.md
```diff
@@ -86,7 +86,7 @@ a new version of it.
 This is not desirable when you have complicated hierarchies of views, especially
 when some of those views may be materialized and take a long time to recreate.
 
-You can use `replace_view` to generate a CREATE OR REPLACE VIEW SQL statement.
+You can use `replace_view`. `replace_view` generates a CREATE OR REPLACE VIEW SQL statement on a normal view. For a materialized view, it creates a new materialized view under a temporary name (including the index definition specified in that version of the materialized view), then it drops the previous existing materialized view and then renames the new materialized view to its original name.
```
calebthompson: wrap at 80 chars to match surrounding code

derekprior:
How does this differ from `REFRESH MATERIALIZED VIEW CONCURRENTLY`?

> Refresh the materialized view without locking out concurrent selects on the materialized view. Without this option a refresh which affects a lot of rows will tend to use fewer resources and complete more quickly, but could block other connections which are trying to read from the materialized view. This option may be faster in cases where a small number of rows are affected.
> 
> This option is only allowed if there is at least one UNIQUE index on the materialized view which uses only column names and includes all rows; that is, it must not index on any expressions nor include a WHERE clause.
> 
> This option may not be used when the materialized view is not already populated.
> 
> Even with this option only one REFRESH at a time may run against any one materialized view.

It seems you might be better off with adding a unique id column and using `CONCURRENTLY`?

richddr:
@calebthompson Addressed your comments. Still can't see the CodeClimate issues.

@derekprior Not exactly, for my use case that wouldn't work. I have a rails model which is backed by view (a materialized view), not a table. One of my requirements was that it needed to support being able to completely update the SQL definition (including indexes) of the materialized view without having downtime. This means the materialized view could have a completely different SELECT statement, indexes.

This differs from `REFRESH MATERIALIZED VIEW CONCURRENTLY` in that when you do the refresh, you'd do the same SELECT statement, in my case that select statement can change.

richddr:
I believe it is failing on the builds for rails 5.0. But I also don't know if this gem works with rails 5 yet?

derekprior:
@richddr Ah, yes. Sorry, I completely misread the issue description. I have indeed followed this pattern before, manually. I will review.

And yes, this gem does support Rails 5.

gillisd:
What's the status on this PR? This feature would be really useful

richddr:
@derekprior what's the status on this?

spec/scenic/statements_spec.rb
```diff
@@ -108,17 +108,6 @@ module Scenic
           .with(:name, definition.to_sql)
       end
 
-      it "fails to replace the materialized view in the database" do
```
calebthompson: Rather than just deleting this, can you change it to describe the new behavior?

lib/scenic/statements.rb
```diff
@@ -110,13 +109,13 @@ def replace_view(name, version: nil, revert_to_version: nil, materialized: false
...
 
-      if materialized
-        raise ArgumentError, "Cannot replace materialized views"
-      end
-
       sql_definition = definition(name, version)
 
-      Scenic.database.replace_view(name, sql_definition)
+      if materialized
+        Scenic.database.replace_materialized_view(name, sql_definition)
```
gillisd: I think 
```
Scenic.database.replace_materialized_view(name, sql_definition)
```
should be
```
Scenic.database.update_materialized_view(name, sql_definition)
```
otherwise `IndexReapplication.new` won't be called

richddr: If you want to re-apply or attempt to re-apply some of the previous  existing indexes you should use `#update_materialized_view` instead.

richddr: If you want to re-apply or attempt to re-apply some of the previous  existing indexes you should use `#update_view` instead.

richddr: See line 86 of this file.

spec/scenic/statements_spec.rb
richddr: @calebthompson At the time I tried to and was running into a few errors that I didn't quite understand and didn't have the time to fix. If this is a blocker for getting this merged I can try again.

calebthompson: Yeah I don't think we can get this in without coverage. Changing the deleted spec to exercise the new path successfully would cover it I think.

calebthompson:
There's the missing test and a hound violation that should be able to be fixed by splitting the sql over a couple of lines, then I think we can get this in the upcoming release.

richddr:
@calebthompson added the test and addressed the hound violation.

richddr:
@calebthompson status on this?

derekprior:
Should this be the default for `update_view` with a materialized view? I'm having a hard time figuring out why you wouldn't want to minimize the time the table is locked?

richddr:
@derekprior I don't understand your question...

With this PR, this is `update_view` with a materialized view:

```

    def update_view(name, version: nil, sql_definition: nil, revert_to_version: nil, materialized: false)
      if version.blank? && sql_definition.blank?
        raise(
          ArgumentError,
          "sql_definition or version must be specified",
        )
      end

      if version.present? && sql_definition.present?
        raise(
          ArgumentError,
          "sql_definition and version cannot both be set",
        )
      end

      sql_definition ||= definition(name, version)

      if materialized
        Scenic.database.update_materialized_view(name, sql_definition)
      else
        Scenic.database.update_view(name, sql_definition)
      end
    end
```

where :

```
      def update_materialized_view(name, sql_definition)

        IndexReapplication.new(connection: connection).on(name) do
          replace_materialized_view(name, sql_definition)
        end
      end
```

which : 
```
      def replace_materialized_view(name, sql_definition)

        temp_mv_name = "#{name}_temp"

        mv_definition = get_definition_for(:create_definition, sql_definition)
        create_materialized_view(temp_mv_name, mv_definition.first)

        new_index_definitions = get_definition_for(:indexes, sql_definition)

        new_index_definitions.each do |idx_definition|
          execute idx_definition.gsub(name, temp_mv_name)
        end

        drop_materialized_view(name)

        sql = "ALTER MATERIALIZED VIEW #{quote_table_name(temp_mv_name)} " +
          "RENAME TO #{quote_table_name(name)};"

        execute sql

        update_index_names_for(temp_mv_name, name)
      end
```

So both `#update_view` and `#replace_view` run the same method underneath (`#replace_materialized_view`). Only difference is that `#update_view` attempts to re-apply the existing indexes and `#replace_view` does not.

derekprior:
Why is that a useful distinction to keep?

richddr:
@derekprior I didn't want to change that existing functionality. Perhaps there are some users out there who share the same opinion as me in that I think of using `#update_view` when you want to make small updates to the existing view. `#replace_view` when you want to completely change the view. That's how I've been using it.

lib/scenic/adapters/postgres.rb
```diff
@@ -153,11 +155,49 @@ def update_materialized_view(name, sql_definition)
...
+      # @return [void]
+      def replace_materialized_view(name, sql_definition)
+        raise_unless_materialized_views_supported
+
+        temp_mv_name = "#{name}_temp"
+
+        mv_definition = get_definition_for(:create_definition, sql_definition)
+        create_materialized_view(temp_mv_name, mv_definition.first)
+
+        new_index_definitions = get_definition_for(:indexes, sql_definition)
```
derekprior: We do not support specifying indexes in the SQL file. I see you've added that here, but I'd prefer to have people use `add_index` in migrations for this. I do not want to duplicate what can already be done with AR migrations.

lib/scenic/adapters/postgres.rb
```diff
@@ -153,11 +155,49 @@ def update_materialized_view(name, sql_definition)
...
+      # @param sql_definition The SQL schema for the updated view.
+      #
+      # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
+      #   in use does not support materialized views.
+      #
+      # @return [void]
+      def replace_materialized_view(name, sql_definition)
+        raise_unless_materialized_views_supported
+
+        temp_mv_name = "#{name}_temp"
```
derekprior: This extension of the name might not be valid depending on the original length of the name. Perhaps we should use a static name that we know to be short. `_scenic_mv` or something?

derekprior:
This general feature is something we'd support but as mentioned, I think this approach is a bit too far-reaching for what we're after.

I'm going to take a shot a simplifying this a bit... hopefully over the next couple of weeks.

mbell697:
@derekprior Has anything progressed with this? If not could you provide a high level description of how you'd prefer it be implemented? This is causing some problems for us and I can spend some time implementing it.

derekprior:
@mbell697 I didn't get very far. It's been a while since I looked through this, but I think I'd be in favor of the approaches here: https://github.com/thoughtbot/scenic/pull/160#pullrequestreview-17703091

I can think of lots of complication here though, which is why I'm hesitant to just say "yes, we will definitely do this" or "yes, go spend hours implementing this". Here are some complications off the top of my head:

1. What if the migration used `disable_ddl_transaction`? It's possible in this scenario that we may be left in a state where the temp view exists but wasn't promoted. Should we disable this feature if we detect DDL transactions are disabled? Should we allow an escape hatch of some sort?

2. The block form I suggested above that would allow you to index the temp view before renaming it might not be a requirement? I'm not really certain. Reapplying indexes can be costly on large views, but that is a cost we already have today. Given that, perhaps this aspect can be a separate PR?

@calebthompson What do you think?

mbell697:
> What if the migration used disable_ddl_transaction?

I think in general the user is taking the risk if disabling ddl transactions and (hopefully) understands the possible consequences of this. Additionally I'd argue that the current code is _more_ dangerous when run with `disable_ddl_transaction` than what is being proposed here. At present you could end up with no table at all, actually I think you're guaranteed to be in a state of the table not existing for some period of time with `disable_ddl_transaction` in the current code. In the approach presented, nothing impacts the in use table other than a single atomic rename operation so even with `disable_ddl_transaction` you are safe in the sense that you'll either be left with a correctly updated production table, or an untouched old version of the table and possible a dangling temp table. 

> The block form I suggested above that would allow you to index the temp view before renaming it might not be a requirement?

I think the current re-indexing strategy is fine, just apply them to the temp table prior to the rename and concurrent index creation isn't needed for this. The block to customize behavior is a reasonable feature, but I see it as a separate issue from the update strategy. 

If there is concern about changing existing behavior, how about making this an optional update strategy set via a param in the migration?  If nothing else, adding some sort of hooks to allow customizing the update strategy would be great. At the moment this issue is really a blocker for using scenic to manage materialized views for applications that can't accept downtime while the migration is running, and since materialized views tend to be expensive to build by nature we're talking possibly many minutes of downtime.

PR #162, opened 2016-07-29
Add support for sql_definition parameter to update_view

jwood:
This pull requests adds support for the `sql_definition` parameter when updating views.

### DIFF
--- a/lib/scenic/statements.rb
+++ b/lib/scenic/statements.rb
@@ -6,9 +6,9 @@ module Scenic
     # @param name [String, Symbol] The name of the database view.
     # @param version [Fixnum] The version number of the view, used to find the
     #   definition file in `db/views`. This defaults to `1` if not provided.
-    # @param sql_definition [String] The SQL query for the view schema. If both
-    #   `sql_defintiion` and `version` are provided, `sql_definition` takes
-    #   prescedence.
+    # @param sql_definition [String] The SQL query for the view schema. An error
+    #   will be raised if `sql_definition` and `version` are both set,
+    #   as they are mutually exclusive.
     # @param materialized [Boolean] Set to true to create a materialized view.
     #   Defaults to false.
     # @return The database response from executing the create statement.
@@ -21,12 +21,16 @@ module Scenic
     #     SELECT * FROM users WHERE users.active = 't'
     #   SQL
     #
-    def create_view(name, version: 1, sql_definition: nil, materialized: false)
-      if version.blank? && sql_definition.nil?
+    def create_view(name, version: nil, sql_definition: nil, materialized: false)
+      if version.present? && sql_definition.present?
         raise(
           ArgumentError,
-          "view_definition or version_number must be specified"
+          "sql_definition and version cannot both be set",
         )
+      end
+
+      if version.blank? && sql_definition.blank?
+        version = 1
       end
 
       sql_definition ||= definition(name, version)
@@ -66,6 +70,9 @@ module Scenic
     #
     # @param name [String, Symbol] The name of the database view.
     # @param version [Fixnum] The version number of the view.
+    # @param sql_definition [String] The SQL query for the view schema. An error
+    #   will be raised if `sql_definition` and `version` are both set,
+    #   as they are mutually exclusive.
     # @param revert_to_version [Fixnum] The version number to rollback to on
     #   `rake db rollback`
     # @param materialized [Boolean] True if updating a materialized view.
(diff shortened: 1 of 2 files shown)


lib/scenic/statements.rb
```diff
@@ -25,7 +25,7 @@ def create_view(name, version: 1, sql_definition: nil, materialized: false)
       if version.blank? && sql_definition.nil?
         raise(
           ArgumentError,
-          "view_definition or version_number must be specified"
+          "sql_definition or version must be specified",
```
calebthompson: The argument still appears to be version_number here.

lib/scenic/statements.rb
```diff
@@ -66,6 +66,9 @@ def drop_view(name, revert_to_version: nil, materialized: false)
     #
     # @param name [String, Symbol] The name of the database view.
     # @param version [Fixnum] The version number of the view.
+    # @param sql_definition [String] The SQL query for the view schema. If both
+    #   `sql_defintiion` and `version` are provided, `sql_definition` takes
```
calebthompson: Typo defintiion

lib/scenic/statements.rb
calebthompson: Never mind I need more coffee

calebthompson:
@jwood just a typo fix, then please squash your commits.

calebthompson:
Failing test seems to be a Travis problem.

jwood:
@calebthompson Typo fixed, and commits squashed.

derekprior:
What's the use case here?

Also, I wouldn't necessarily worry about the breaking change. As far as I understand it, it's a bug we are fixing.

calebthompson:
@derekprior my understanding is that this will bring us into parity with create_view's sql_definition option, which makes sense to me.

jwood:
The use case is that we need to update a view that we used `create_view` to create, using `create_view`'s sql_definition option. We have a SQL ERB template that we first feed through ERB, and then feed the generated SQL to `create_view`/`update_view`.

jwood:
Hi @calebthompson and @derekprior. Thanks for all the feedback. I'm just looking to get a little clarification on the change @calebthompson suggested here: https://github.com/thoughtbot/scenic/pull/162#commitcomment-18464891

I'm more than happy to make the change, but if we do, I think we should change both `create_view` and `update_view` for consistency. And, that would be a backwards incompatible change for `create_view` (not for `update_view`, because that would be new functionality).

Please let me know if you would like me to raise an error if both the `version` and the `sql_definition` options are provided.

jwood:
Hi @calebthompson and @derekprior. Any thoughts on how to push this through?

calebthompson:
> Please let me know if you would like me to raise an error if both the version and the sql_definition options are provided.

We would, and I think it's a bug that we didn't before and instead failed silently, so it's not backward-incompatible.

lib/scenic/statements.rb
```diff
@@ -6,9 +6,9 @@ module Statements
     # @param name [String, Symbol] The name of the database view.
     # @param version [Fixnum] The version number of the view, used to find the
     #   definition file in `db/views`. This defaults to `1` if not provided.
-    # @param sql_definition [String] The SQL query for the view schema. If both
-    #   `sql_defintiion` and `version` are provided, `sql_definition` takes
-    #   prescedence.
+    # @param sql_definition [String] The SQL query for the view schema. An error
+    #   will be raised if `sql_defintiion` and `version` are both set,
```
calebthompson: Do you mind fixing the typo here as well?

lib/scenic/statements.rb
```diff
@@ -75,12 +78,15 @@ def drop_view(name, revert_to_version: nil, materialized: false)
     # @example
     #   update_view :engagement_reports, version: 3, revert_to_version: 2
     #
-    def update_view(name, version: nil, revert_to_version: nil, materialized: false)
-      if version.blank?
-        raise ArgumentError, "version is required"
+    def update_view(name, version: nil, sql_definition: nil, revert_to_version: nil, materialized: false)
```
calebthompson: This is fine.

jwood:
OK @calebthompson. I've added code to raise an error if both `version` and `sql_version` are specified in calls to `create_view` or `update_view`.

Sorry about the extra commits to fix the hound violations. I'll squash them if everything else looks OK.

calebthompson:
Couple of failing tests with the new error. Can you take a look?

calebthompson:
Interesting, the version being omitted caused the error?

jwood:
Yes, that's the case. Because, an omitted version will default to "1". So, we end up with a version of "1" and a defined `sql_definition`...which triggers the error.

calebthompson:
And we have the default `version: 1` in `Scenic::Statements#create_view` to allow `create_view :viewname` for the initial version.

Unfortunately I think that does make it a breaking change. We could possibly default `version` if both `version` and `sql_definition` are omitted (rather than the error we raise now), which seems slightly less bad to me.

@derekprior What do you think? The options are:

1. Leave this as-is, which will break if someone is currently doing `create_view :viewname, sql_definition: ...` because version is defaulted
2. Default `version` to 1 somewhere in the method body if neither `version` nor `sql_definition` are provided (safer, since this is an error state now)
3. Don't raise an error if both are provided, but let sql_definition override version number. Maybe omit a warning in this case as well.

My vote is for 2., as it's turning an error case into non-error (therefore not breaking) and allows both

```ruby
create_view :viewname
create_view :otherviewname, sql_definition: ...
```

calebthompson:
(sorry to flip-flop, @jwood)

jwood:
@calebthompson I like your suggestion of defaulting the version in the method body. Let me push up something along those lines.

calebthompson:
LGTM. Please squash this and I'll merge.

jwood:
@calebthompson Commits have been squashed

PR #167, opened 2016-09-15
Updated the readme to provide information about updating the materialized views

symmetrically:
The readme lacked information on updating the materialized view. Updated the readme to provide more information on how to update the views when they are materialized.


richddr:
Hopefully after https://github.com/thoughtbot/scenic/pull/160 goes in this won't be needed.

derekprior:
I think we could add a note where we discuss `replace_view` to say it is not supported for materialized views. As is, I think this is a bit too specific for the goals of the README.

PR #168, opened 2016-09-21
Add 1.3.0 to changelog

jaredbeck:


derekprior:
Thanks. I revised this to be in line with our release notes on GitHub's releases UI and merged.

jaredbeck:
Thanks Derek. Just out of curiosity, why do you prefer GH releases over plain text. I like plain text because it'll still work the same in 20 years and is easy to search, but maybe there's a killer feature of GH releases that I'm missing? Just curious, thanks.

derekprior:
I prefer the plaintext `NEWS` myself. GitHub releases share some things with annotated tags, which makes them easier to create. I think that's why 1.3.0 ended up with entries in GH releases but not in NEWS.

PR #169, opened 2016-09-30
Add Ability to maintain functions

paulccarey:
This adds the ability to manage functions in much the same way as Scenic already handles views.

I have maintained the architecture and structure of Scenic, splitting functionality between views and functions where appropriate.

Functions are dumped out into `schema.rb` and functions are maintained in `db/functions`.

Function generators have been added also as well as the usual array of `create_function`, `drop_function`,  etc methods.


teoljungberg:
After a conversation with @derekprior and @calebthompson about this a while back, we felt that it was better for Scenic to only be responsible for views and not anything else database related. To keep the tool itself small and focused.

I then started to write [F(x)] which takes the same approach as Scenic does, but focuses on database triggers and functions.

Try it out, and please open an issue if you encounter any issues!

[F(x)]: https://github.com/teoljungberg/fx

derekprior:
As @teoljungberg mentioned, F(x) seems like a good option here.

PR #171, opened 2016-10-28
Add support for `rails@3.2`

seanpdoyle:
This commit relaxes the `activerecord` and `activesupport` version
requirements from `rails@4.0` to `rails@3.2`.


seanpdoyle:
I spoke with @derekprior about adding support for versions of Rails that are [past their end of life](http://guides.rubyonrails.org/maintenance_policy.html).

We decided that it'd be best to stick with supporting Rails 4+.

PR #172, opened 2016-11-08
Support for unusual characters on a view's name

aokiji:
I have to support latin1 characters on some view's name, so I made a few fixes so that I could still use scenic.

### DIFF
--- a/lib/scenic/adapters/postgres/views.rb
+++ b/lib/scenic/adapters/postgres/views.rb
@@ -43,9 +43,9 @@ module Scenic
           namespace, viewname = result.values_at "namespace", "viewname"
 
           if namespace != "public"
-            namespaced_viewname = "#{namespace}.#{viewname}"
+            namespaced_viewname = "#{pg_identifier(namespace)}.#{pg_identifier(viewname)}"
           else
-            namespaced_viewname = viewname
+            namespaced_viewname = pg_identifier(viewname)
           end
 
           Scenic::View.new(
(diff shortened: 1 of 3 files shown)


lib/scenic/adapters/postgres/views.rb
```diff
@@ -54,6 +54,13 @@ def to_scenic_view(result)
             materialized: result["kind"] == "m",
           )
         end
+
+        private
+
+        def pg_identifier(name)
+          return name if name =~ /^[a-zA-Z_][a-zA-Z0-9_]*$/
+          PGconn.quote_ident(name)
```
derekprior: Can we always quote the name? Or will that not work?

spec/scenic/schema_dumper_spec.rb
```diff
@@ -1,6 +1,7 @@
 require "spec_helper"
 
 class Search < ActiveRecord::Base; end
+class SearchInAHaystack < ActiveRecord::Base; self.table_name = '"search in a haystack"'; end
```
calebthompson: This can be split to three lines.

calebthompson:
@aokiji I deleted a few pieces of Hound feedback I'm comfortable ignoring here, but if you don't mind taking a look at the rest we can try and get this out in the upcoming release. It's mostly whitespace changes and splitting a few things over multiple lines.

calebthompson:
If you rebase onto master, it may clear up some of the CI issues as well.

aokiji:
I rebased and acted on some of the feedback from Hound. Now all tests are passing. @calebthompson I would be very glad to see this change on the next release

calebthompson:
Merged in #188

PR #173, opened 2016-11-14
Fix deprecation warnings in test code

devonestes:
There were a bunch of rather verbose deprecation warnings in the test output,
which I fixed by adding a check for Rails 5 to ensure that we use valid syntax
for that version while also keeping backwards compatibility for earlier versions.


teoljungberg:
I made this change to [Fx] too and it works fine and dandy.


[Fx]: https://github.com/teoljungberg/fx

derekprior:
THanks @devonestes !

PR #174, opened 2016-11-14
Add refreshing materialized views with dependencies

devonestes:
We had issues with refreshing complex chains of materialized views that
depended on one another. For example, we had the following type of thing
happening.

(imagine all the relations below are materialized views)

A depends on B
B depends on C and D
C depends on D
D depends only on non-materialized views

Now if you wanted to refresh A, and you wanted it to be as fresh as possible,
first you would need to refresh the materialized views on which it depends,
followed by A itself. This became frequent enough, and changed with enough
frequency, to make keeping refresh methods for each individual materialized
view unreasonable.

I realized we could use the information kept in `pg_depend` for this purpose,
and also thought that I'm most likely not the first person to have this
problem, so I'm contributing it back here. Also, I can see further improvements
to the migrations being built off of this usage of the dependency graph stored
in Postgres, so I thought it might help there, too!

There's more refactoring that can be done here, but I wanted to get this
working (but imperfect) version up to get feedback on design, style, etc.

Also, how much, and where, would you like this documented?

cc: @calebthompson

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -3,6 +3,7 @@ require_relative "postgres/errors"
 require_relative "postgres/index_reapplication"
 require_relative "postgres/indexes"
 require_relative "postgres/views"
+require_relative "postgres/refresh_dependencies"
 
 module Scenic
   # Scenic database adapters.
@@ -195,8 +196,9 @@ module Scenic
       #   Scenic.database.refresh_materialized_view(:posts, concurrent: true)
       #
       # @return [void]
-      def refresh_materialized_view(name, concurrently: false)
+      def refresh_materialized_view(name, concurrently: false, cascade: false)
         raise_unless_materialized_views_supported
+        refresh_dependencies_for(name) if cascade
 
         if concurrently
           raise_unless_concurrent_refresh_supported
@@ -225,6 +227,14 @@ module Scenic
         unless connection.supports_concurrent_refreshes?
           raise ConcurrentRefreshesNotSupportedError
         end
+      end
+
+      def refresh_dependencies_for(name)
+        Scenic::Adapters::Postgres::RefreshDependencies.call(
+          name,
+          self,
+          connection,
+        )
       end
     end
   end
--- a/lib/scenic/adapters/postgres/refresh_dependencies.rb
+++ b/lib/scenic/adapters/postgres/refresh_dependencies.rb
@@ -1,0 +1,67 @@
+module Scenic
+  module Adapters
+    class Postgres
+      class RefreshDependencies
+        def self.call(name, adapter, connection)
+          new(name, adapter, connection).call
+        end
+
+        def initialize(name, adapter, connection)
+          @name = name
+          @adapter = adapter
+          @connection = connection
+        end
+
+        def call
+          dependencies.each do |dependency|
+            adapter.refresh_materialized_view(dependency)
+          end
+        end
+
+        private
+
+        attr_reader :name, :adapter, :connection
+
+        DEPENDENCY_SQL = <<-SQL.freeze
+          SELECT r_ns.nspname || '.' || cl_r.relname AS materialized_view,
+          array_agg(d_ns.nspname || '.' || cl_d.relname) AS depends_on
+          FROM pg_rewrite AS r
+          JOIN pg_class AS cl_r ON r.ev_class=cl_r.oid
+          JOIN pg_depend AS d ON r.oid=d.objid
+          JOIN pg_class AS cl_d ON d.refobjid=cl_d.oid
+          JOIN pg_namespace AS r_ns ON r_ns.oid = cl_r.relnamespace
+          JOIN pg_namespace AS d_ns ON d_ns.oid = cl_d.relnamespace
+          WHERE cl_d.relkind = 'm'
+          AND cl_r.relkind = 'm'
+          AND cl_d.relname != cl_r.relname
+          GROUP BY cl_r.relname, r_ns.nspname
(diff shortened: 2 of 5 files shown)


lib/scenic/adapters/postgres/refresh_dependencies.rb
```diff
@@ -0,0 +1,67 @@
...
+          dependencies.each do |dependency|
+            adapter.refresh_materialized_view(dependency)
+          end
+        end
+
+        private
+
+        attr_reader :name, :adapter, :connection
+
+        DEPENDENCY_SQL = <<~SQL.freeze
```
calebthompson: Looks like the `<<~` construct is tripping up CodeClimate. It's using a 2.1 ruby parser for rubocop and gives the error "unexpected token tLSHFT". The other two issues it notes are because the SQL heredoc is being interpreted as code instead of a string. Can you look into that?

derekprior:
I like this idea. Would `cascade` make sense as the option name? Is this something we should always do or should it be opt-in as it is here?

What about the code we generate in models for materialized views?

lib/scenic/adapters/postgres/refresh_dependencies.rb
calebthompson: Same thing in Travis: https://travis-ci.org/thoughtbot/scenic/jobs/175636690#L267

lib/scenic/adapters/postgres.rb
```diff
@@ -195,8 +196,10 @@ def drop_materialized_view(name)
       #   Scenic.database.refresh_materialized_view(:posts, concurrent: true)
       #
       # @return [void]
-      def refresh_materialized_view(name, concurrently: false)
+      def refresh_materialized_view(name, concurrently: false,
+                                    with_dependencies: false)
```
calebthompson: I prefer to have signatures on one line, even if they go longer than whatever length, for method definitions. Can you join these?

lib/scenic/adapters/postgres.rb
```diff
@@ -226,6 +229,11 @@ def raise_unless_concurrent_refresh_supported
           raise ConcurrentRefreshesNotSupportedError
         end
       end
+
+      def refresh_dependencies_for(name)
+        Scenic::Adapters::Postgres::RefreshDependencies.call(name, self,
+                                                             connection)
```
calebthompson: We like to do one arg per line, and no args on lines with parens, so

```
Scenic::Adapters::Postgres::RefreshDependencies.call(
  name,
  self,
  connection,
)
```

devonestes:
Ok, I made the changes requested by @calebthompson, and I liked the name `cascade` way better, so I made that change as well!

I think that by default we should have `cascade: false` since that's not what I would expect as the behavior of that method by default, so I think that's the least surprising thing that we can do. I like putting the option expressly in the model code from the generators, though, so that users know that's available to them, so I've gone and done that as well.

lib/scenic/adapters/postgres/refresh_dependencies.rb
```diff
@@ -0,0 +1,67 @@
...
+          dependencies.each do |dependency|
+            adapter.refresh_materialized_view(dependency)
+          end
+        end
+
+        private
+
+        attr_reader :name, :adapter, :connection
+
+        DEPENDENCY_SQL = <<-SQL.freeze
```
derekprior: If the intention is for this to be private then it should be followed with a `private_constant "DEPENDENCY_SQL"`

lib/scenic/adapters/postgres/refresh_dependencies.rb
```diff
@@ -0,0 +1,67 @@
...
+          SELECT r_ns.nspname || '.' || cl_r.relname AS materialized_view,
+          array_agg(d_ns.nspname || '.' || cl_d.relname) AS depends_on
+          FROM pg_rewrite AS r
+          JOIN pg_class AS cl_r ON r.ev_class=cl_r.oid
```
derekprior: I'm having a hard time following the single letter and abreviated aliases here. Can we give them longer, more descriptive names to help maintenance going forward?

calebthompson:
On the one hand, this seems like good default behavior. On the other, it is more understandable to follow what Postgres does and not cascade automatically. 👍

lib/scenic/adapters/postgres/refresh_dependencies.rb
```diff
@@ -0,0 +1,67 @@
...
+          WHERE cl_d.relkind = 'm'
+          AND cl_r.relkind = 'm'
+          AND cl_d.relname != cl_r.relname
+          GROUP BY cl_r.relname, r_ns.nspname
+          ORDER BY cl_r.relname;
+        SQL
+
+        def dependencies
+          dependency_rows = connection.select_rows(DEPENDENCY_SQL)
+          dependency_hash = parse_to_hash(dependency_rows)
```
derekprior: I have a really hard time following what's going on from this point on. Do you think it might make sense to create some objects that represent these rows? They could be private internal classes if desired?

lib/scenic/adapters/postgres.rb
```diff
@@ -195,8 +196,9 @@ def drop_materialized_view(name)
       #   Scenic.database.refresh_materialized_view(:posts, concurrent: true)
       #
       # @return [void]
-      def refresh_materialized_view(name, concurrently: false)
+      def refresh_materialized_view(name, concurrently: false, cascade: false)
         raise_unless_materialized_views_supported
+        refresh_dependencies_for(name) if cascade
```
derekprior: We like to avoid inline conditionals, preferring the block form in most cases. Could you make that change here please?

PR #175, opened 2016-11-15
Update rails50 Appraisals to use Rails 5 proper

teoljungberg:
With the release of Rails 5, there is no need to fetch Rails from its
master branch to install it.


teoljungberg:
With this change, would making a `rails-head` Appraisal make sense to still test against Rails master?

derekprior:
If you want to add rails-head and add it to allowed-failures on travis, I think that'd be okay.

PR #178, opened 2016-11-15
Appraise Rails edge

teoljungberg:
Update Travis build matrix to allow failures if one should occur.


teoljungberg:
References https://github.com/thoughtbot/scenic/pull/175#issuecomment-260666660

derekprior:
Was RSpec master required for this or was it something you thought best for this long term?

teoljungberg:
It's not required, I honestly copied the old `rails50` appraisal (before #175 was merged) where it was required. I don't see any harm in keeping it as-is as a safe-guard for the future, but I can also see the argument that it's a premature change.

What do you think?

calebthompson:
I'm sure @samphippen and friends will appreciate it.

calebthompson:
This is good to go. We can drop the rspec master change later if we need to.

Issue #176, opened 2016-11-15
Generated migrations should inherit from versioned base class in Rails 5


derekprior:
Migrations are currently generated with the base class of `ActiveRecord::Migration`. If we're running under rails 5, we should generate them with `ActiveRecord::Migration[5.0]` to save our users from deprecation warnings.

It's not clear if we should hard code the version check and version number or if we should use something like:

```ruby
def migration_version
  if ActiveRecord::Migration.respond_to?("[]")
    "[#{Rails.version.major}.#{Rails.version.minor}]"
  end
end
```

I'm not sure if they intend to bump this with every version or what...

calebthompson:
I'd be interested in seeing if we can get this automatically. @sgrif do you know if there's an API for this we could use?

sgrif:
`ActiveRecord::Migration.current_version` is the public API to get the generated value.

sgrif:
I think the best course of action would be to specifically target 5.0, and not the current version though.

derekprior:
Why do you say that?

sgrif:
Well, the point of the versioned migrations is that it makes us more free to change the migration API without breaking old migrations. You wrote the generator to generate migrations that are expected to work with the DSL as of 5.0. There's no guarantee that they'll continue to work with future versions, and there's no downside to being specific. (Yes, `execute` is unlikely to ever change, but in the more general sense...)

teoljungberg:
I opened #179 to remedy this issue.

teoljungberg:
This can be closed now since #179 is merged.

PR #179, opened 2016-11-15
Generate migrations warning free on Rails 5

teoljungberg:
Rails 5 introduced versioned migrations to cope with breaking changes in
the migrations API. To remedy those breakages, versioned migrations were
introduced.

The new migration syntax is only generated on Rails 5.0 and above.

References #176

### DIFF
--- a/lib/generators/scenic/view/templates/db/migrate/create_view.erb
+++ b/lib/generators/scenic/view/templates/db/migrate/create_view.erb
@@ -1,4 +1,4 @@
-class <%= migration_class_name %> < ActiveRecord::Migration
+class <%= migration_class_name %> < <%= activerecord_migration_class %>
   def change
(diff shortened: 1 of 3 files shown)


spec/generators/scenic/view/view_generator_spec.rb
```diff
@@ -48,4 +48,19 @@
       expect(view_definition).to exist
     end
   end
+
+  if Rails.gem_version >= Gem::Version.new("5.0")
```
derekprior: I'm not convinced we need to keep this test. This test passing when it was written + ongoing passing of the test suite seems sufficient.

derekprior: In any event, it seems `Rails.gem_version` is not defined in Rails 4.0

PR #180, opened 2016-11-16
Address further comments in cascading materialized view refresh

devonestes:
@derekprior had a few other comments about this feature that didn't get
addressed before it was merged, so this PR is to address those issues. It aims
to make the process of parsing the dependency information from Postgres a
little easier to understand, addresses some style issues, and also adds
documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ refreshes:
 
 ```ruby
 def self.refresh
-  Scenic.database.refresh_materialized_view(table_name, concurrently: false)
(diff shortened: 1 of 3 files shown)


devonestes:
This addresses the comments that weren't addressed in #174. I feel like I still have a ways to go on making the private class `DependencyParser` easier to follow, but for now I wanted to get this up to see if anyone else had feedback or info on maybe what could make that all easier to understand. I'll sleep on it and come back to it tomorrow and hopefully have some better ideas!

PR #182, opened 2016-11-21
Update bin/setup and acceptance_helper.rb

### LINKED ISSUE #177: Acceptance suite cleanup fails locally under Rails 5
The `acceptance_helper.rb` file speciefies that the dummy database should be dropped and recreated after the acceptance suite is run. Rails 5 added some protections to `rake db:drop` so that it will not drop a database that is not explicitly marked as the test database.

When I run my tests locally, this fails because the databases as created in `bin/setup` lack this information, I think.  

This works fine in travis, probably because RAILS_ENV is always equal to "test"? Not clear.

devonestes:
Some folks were encoutering issues with the additional security layer for
databases that came in Rails 5, which was preventing the dropping of the test
database. This commit adds in some conditional stuff to do that additional
setup for Rails 5.

@derekprior - for some reason I wasn't able to see this fail on my machine, but
I'm curious if it does the job for you. I was able to do the five things you
listed in the issue without problem, so hopefully it'll work!

Resolves #177

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -10,3 +10,10 @@ if [ -z "$CI" ]; then
 fi
 
 bundle exec rake dummy:db:create
+
+if ! [ -z "$CI" ]; then
+  rails -v | grep -q '5.'
+  if [[ $? -eq 0 ]]; then
+    bundle exec rake dummy:db:environment:set RAILS_ENV=test
+  fi
+fi
--- a/spec/acceptance_helper.rb
+++ b/spec/acceptance_helper.rb
@@ -21,12 +21,21 @@ RSpec.configure do |config|
 
   config.after(:suite) do
     Dir.chdir("spec/dummy") do
-      system <<-CMD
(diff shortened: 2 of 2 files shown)


bin/setup
```diff
@@ -10,3 +10,9 @@ if [ -z "$CI" ]; then
 fi
 
 bundle exec rake dummy:db:create
+
+rails -v | grep -q '5.0'
+greprc=$?
+if [[ $greprc -eq 0 ]]; then
```
teoljungberg: I would inline `$?`, which seems to be the conventional way.

devonestes: @teoljungberg Yeah - if you haven't noticed, shell scripting isn't my strong point... Thanks for the tip!

derekprior:
I think I finally figured out the various failures I was seeing. The environment was part of, but not the whole, issue. I fixed the various issues in #185.

PR #183, opened 2016-12-19
Sort views by name when dumped to the schema

Dorian:
- Makes for more consistent schema.rb generation
- Makes it easier to spot the changes / diff from the previous version


derekprior:
We originally sorted these by name, but this does not work the minute you have a view that depends on another view. The current sorting of views has been arrived at after several iterations, but here's the original commit that removed the alphabetical ordering: https://github.com/thoughtbot/scenic/commit/68747d5e9f2253a6cb66ce1d56a0f394d81a4d9a

Dorian:
@derekprior Thanks, I didn't consider that as none of my views are inter-dependent.

I didn't see the Tsort that is going on: https://github.com/thoughtbot/scenic/blob/13709f40387da39ebf6004e6031d3ae7e858a925/lib/scenic/adapters/postgres/refresh_dependencies.rb#L31-L53

Will try to fix it

PR #185, opened 2017-01-15
Fix test suite issues

derekprior:
* Intermittent failures were caused because one acceptance spec was not
cleaning up after itself.
* Some developers were seeing test suite failures under Rails 5 because
`rake db:environment:set` needed to be run. This task doesn't exist in
earlier versions of Rails, so we created a dummy task in the dummy app
so we could call it nonetheless.
* Edge rails appraisal doesn't need edge rspec

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -21,11 +21,6 @@ if RUBY_VERSION > "2.2.0"
 
   appraise "rails-edge" do
     gem "rails", github: "rails/rails"
-    gem "rspec-rails", github: "rspec/rspec-rails"
-    gem "rspec-support", github: "rspec/rspec-support"
-    gem "rspec-core", github: "rspec/rspec-core"
-    gem "rspec-mocks", github: "rspec/rspec-mocks"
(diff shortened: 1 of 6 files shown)


Appraisals
```diff
@@ -21,11 +21,6 @@ if RUBY_VERSION > "2.2.0"
 
   appraise "rails-edge" do
     gem "rails", github: "rails/rails"
-    gem "rspec-rails", github: "rspec/rspec-rails"
-    gem "rspec-support", github: "rspec/rspec-support"
-    gem "rspec-core", github: "rspec/rspec-core"
-    gem "rspec-mocks", github: "rspec/rspec-mocks"
-    gem "rspec-expectations", github: "rspec/rspec-expectations"
-    gem "rspec", github: "rspec/rspec"
+    gem "arel", github: "rails/arel"
```
teoljungberg: Why is arel's master needed? There's not mention of it in the PR description.

derekprior: It was required to fulfill edge rails' version constraints

PR #188, opened 2017-02-07
Quote view names

calebthompson:


calebthompson:
Closes #172

lib/scenic/adapters/postgres/views.rb
```diff
@@ -54,6 +54,11 @@ def to_scenic_view(result)
             materialized: result["kind"] == "m",
           )
         end
+
+        def pg_identifier(name)
+          return name if name =~ /^[a-zA-Z_][a-zA-Z0-9_]*$/
```
derekprior: Can we not safely quote the identifier in any case?

calebthompson:
I tried removing this and it ended up causing double quoting because we'd quote on the way out of the db and when it goes in. I think it's only code, but it does break the scheme dump.

lib/scenic/adapters/postgres/views.rb
calebthompson: Removing this line caused double quoting in in-memory representations of the names (failing tests) and in the SQL dump (failing more tests, and breaking things). I wasn't able to fix that in ~an hour, so I reverted my attempts.

PR #189, opened 2017-02-09
Allow for / in view/table name

voltechs:
Did a bit of refactoring to allow a "/" in the view name.


derekprior:
Is this potentially all covered by #188?

calebthompson:
Should be.

calebthompson:
Resolved by #188

PR #191, opened 2017-02-15
WIP Sorting views

Dorian:
Sorts alphabetically nodes of dependencies graph.

- [ ] fix specs
- [ ] add specs for more complex dependency trees
- [ ] is the `sort` in `each_node` needed if views are already sorted when fetched?
- [ ] why were the views sorted by `iod`?


Dorian:
From #183, commit that removed the ordering: https://github.com/thoughtbot/scenic/commit/68747d5e9f2253a6cb66ce1d56a0f394d81a4d9a

TSort (topological sorting for dependencies) was added a year later, e.g. `RefreshDependencies`, but can be easily adapted to sorted by name now.

spec/scenic/schema_dumper_spec.rb
```diff
@@ -36,4 +36,48 @@ class Search < ActiveRecord::Base; end
...
+      ActiveRecord::SchemaDumper.dump(Search.connection, stream)
+
+      output = stream.string
+
+      views.each do |name, _|
+        expect(output).to include "create_view :'#{name}',"
+      end
+
+      create_views = output.lines.grep(/create_view :/)
+      order = create_views.map { |line| line.scan(/scenic\.[a-z_0-9]+/).first }
```
calebthompson: `\w` covers the group you want: `/\w/ - A word character ([a-zA-Z0-9_])` So I think this can be `/scenic\.\w+/`.

spec/scenic/schema_dumper_spec.rb
```diff
@@ -36,4 +36,48 @@ class Search < ActiveRecord::Base; end
       Search.connection.drop_view :'scenic.searches'
     end
   end
+
+  context "with dependent views " do
+    it "dumps create_view sorted alphabetically" do
+      Search.connection.execute("CREATE SCHEMA scenic")
+      Search.connection.execute("SET search_path TO scenic, public")
```
calebthompson: These aren't undone after this test, so I think you've got a leaky test here. Do we need the scenic schema?

spec/scenic/schema_dumper_spec.rb
```diff
@@ -36,4 +36,48 @@ class Search < ActiveRecord::Base; end
...
+      end
+
+      stream = StringIO.new
+
+      ActiveRecord::SchemaDumper.dump(Search.connection, stream)
+
+      output = stream.string
+
+      views.each do |name, _|
+        expect(output).to include "create_view :'#{name}',"
```
calebthompson: Given the `create_views.map` lines, I think we don't need these assertions.

spec/scenic/schema_dumper_spec.rb
```diff
@@ -36,4 +36,48 @@ class Search < ActiveRecord::Base; end
...
+        "scenic.view1_1": "SELECT * FROM scenic.view1",
+        "scenic.view1_2": "SELECT * FROM scenic.view1",
+      }
+
+      expected_order = %w(
+        scenic.view1
+        scenic.view1_1
+        scenic.view1_2
+        scenic.view2
+      )
```
calebthompson: This is the same order you'd expect from an alphabetical sort. You cover this a bit in postgres_spec but...

Can you complicate things by:

1. Adding something that relies on scenic.view1 and scenic.view1_1
2. Changing some names so that the expected ordering is non-alphabetical

spec/scenic/schema_dumper_spec.rb
Dorian: Thanks, I removed the scenic namespace

Dorian:
Ok, I figured out the topological sorting part, almost there :) http://stackoverflow.com/questions/42917771/alphabetical-topological-sorting-in-ruby

hcarver:
We're hitting this bug too - anything I can help with to help get it fixed?

Dorian:
@hcarver Yeah, we could try fixing it together if you wanna pair? My email is dorian@doma.io (I live in San Francisco)

Dorian:
@hcarver I just pushed my latest change (a more complex spec basically)

hcarver:
Hi @Dorian, I'm in London so pairing might be tricky! But if you tell me what your planned approach is, what you've done and what's left to do, I'll have a go at helping out.

simonc:
Hi, we're facing the same issue with my team and I'd be glad if I could help with issue. I know it's been almost a year but is anyone still working on this? Thanks ❤️

simonc:
In the mean time, if you're using basic views with no inter-dependencies you can use this quick and (very) dirty override:

```ruby
# lib/scenic_override.rb
module ScenicOverride
  private def dumpable_views_in_database
    super.sort_by(&:name)
  end
end

module Scenic
  module SchemaDumper
    prepend ScenicOverride
  end
end
```
```ruby
# config/application.rb
require_relative '../lib/scenic_override'
```

cspherical:
Our team uses and loves this gem, although this reordering is really becoming a pain point.  Is there anything we can do to help get this merged and released?

derekprior:
I think what remains to be done is addressing code review feedback and then possibly creating more complex test cases that prove this behaves as expected. @cspherical if you want to take another crack at this, either based on this or from scratch, go for it. I just haven't had the time to drive it, but I would review it.

simonc:
Hello there 👋 

Is there anything we can do to help this PR to move forward? This issue is causing a lot of pain in our team so we're happy to help! 🙂 

Cheers.

calebthompson:
@derekprior thoughts on cutting a release if we have unreleased commits, then merging this to master for folks to try as a “beta”?

mcmire:
We are also seeing an issue where views are not being sorted topologically in `schema.rb` and it is causing pain. However, looking over this PR, this doesn't seem to solve that problem exactly. There are two changes here: one to the query that Scenic uses to get all views so it knows what to put in `schema.rb`, and another to the RefreshDependencies class. For the first change, views are merely being sorted alphabetically now instead of when they were added, and I'm not sure how that will help things. In the second case, it does not seem that RefreshDependencies will ever get called automatically when generating `schema.rb` (but I would appreciate clarity around that as I'm not certain about that).

In order to fix `schema.rb`, then, it seems that we'd actually want to modify [Scenic::Adapter::Postgres::Views](https://github.com/scenic-views/scenic/blob/master/lib/scenic/adapters/postgres/views.rb) so that its `all` method returns a topologically sorted array of views. In other words, pull out the tsort code from RefreshDependencies and apply it here somehow. Am I close here?

spec/scenic/schema_dumper_spec.rb
```diff
@@ -36,4 +36,50 @@ class Search < ActiveRecord::Base; end
...
+      output = stream.string
+
+      views.each do |name, _|
+        expect(output).to include "create_view :#{name},"
+      end
+
+      create_views = output.lines.grep(/create_view :/)
+      order = create_views.map { |line| line.scan(/:([a-z_0-9]+),/).first.first }
+
+      expect(order).to eq(expected_order)
```
calebthompson: @mcmire isn't this testing the behavior you describe [in your comment](https://github.com/scenic-views/scenic/pull/191#issuecomment-498364427)?

mcmire: It is, that's true, although I'm struggling to see how, since all this PR seems to do is change how views are being sorted when they're spit out (other than the materialized view stuff, but it seems like that's separate?), and it changes it so it's no longer an alphabetical sort but a sort on `oid`. I'm not 100% certain what that is, I'm guessing it corresponds to the order in which the view was added initially, but there are some question marks here for sure. I'll have to dive into this further.

PR #192, opened 2017-03-06
Added installation commands in readme

tarunjadhwani:
**What?**
Updated github readme file with gem installation instructions

**Why?**
Makes it easy for the user

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,18 @@ Scenic ships with support for PostgreSQL. The adapter is configurable (see
 `Scenic::Configuration`) and has a minimal interface (see
 `Scenic::Adapters::Postgres`) that other gems can provide.
 
+## Installation
+
+Include scenic as a gem in your Gemfile:
+```ruby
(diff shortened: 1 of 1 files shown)


calebthompson:
Thank you for the contribution, but I think at this point enough folks know how to install gems that instructions aren't necessary for these basic cases.

tarunjadhwani:
Agreed that users know how to install a gem, but its good to convey to the users which is the stable version and can be used in production. All gems like [paperclip](https://github.com/thoughtbot/paperclip), [capistrano](https://github.com/capistrano/capistrano), [devise](https://github.com/plataformatec/devise) still add installation updating the gem version with the latest stable version in readme. Its a good to have, but surely your choice :)

allthesignals:
Yeah, I like to know that I'm installing the right gem. It's one of the first things I look for when I want to use a gem. Guessing that the rubygems name is "scenic" doesn't inspire a lot of confidence - could easily be a name for something else. Thanks!

PR #193, opened 2017-03-17
Don’t create a factory when using the model generator

tjmw:
I can't think of a scenario where generating a factory for a scenic view backed model would be useful, so skip it when using the model generator.

I'm not sure if the way I'm testing this is optimal and would definitely welcome some feedback. I've added a new appraisal which includes `rspec-rails` and `factory_girl_rails`, to simulate the conditions under which a factory would be generated. The spec that I've added would fail for this appraisal without the change to `Scenic::Generators::ModelGenerator`. It would not fail for any of the other appraisals because we'd never attempt to generate a factory in these cases.


spec/generators/scenic/model/model_generator_spec.rb
```diff
@@ -32,5 +32,12 @@ module Scenic::Generators
       expect(model_definition).to contain("self.refresh")
       expect(model_definition).to have_correct_syntax
     end
+
+    it "does not create a factory definition" do
+      run_generator ["current_customer"]
+      factory_definition = file("spec/factories/current_customers.rb")
+
+      expect(factory_definition).to_not exist
```
teoljungberg: The thoughtbot styleguide is not to use "to not". As per: https://github.com/thoughtbot/guides/blob/master/style/testing/README.md

tjmw: 👍

PR #196, opened 2017-04-02
Ignore Rails' internal tables while dumping the schema

### LINKED ISSUE #194: Scenic causes `ar_internal_metadata` table to be dumped to schema.rb
I added scenic to my first project with Rails 5 today and noticed this appeared in the schema dump:

```diff
+  create_table "ar_internal_metadata", primary_key: "key", id: :string, force: :cascade do |t|
+    t.string   "value"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
```

The project has been running for 6+ months without this table appearing in the schema, so it seems likely caused by Scenic. It seems harmless, but probably not something we should be dumping.

teoljungberg:
These tables are ignored by Rails itself, and Scenic should do so too.

Resolves #194

### DIFF
--- a/spec/scenic/schema_dumper_spec.rb
+++ b/spec/scenic/schema_dumper_spec.rb
@@ -36,4 +36,18 @@ describe Scenic::SchemaDumper, :db do
       Search.connection.drop_view :'scenic.searches'
     end
   end
+
+  it "ignores tables internal to Rails" do
+    view_definition = "SELECT 'needle'::text AS haystack"
+    Search.connection.create_view :searches, sql_definition: view_definition
+    stream = StringIO.new
+
+    ActiveRecord::SchemaDumper.dump(Search.connection, stream)
+
+    output = stream.string
+
+    expect(output).to include "create_view :searches"
+    expect(output).not_to include "ar_internal_metadata"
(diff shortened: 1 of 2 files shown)


spec/scenic/schema_dumper_spec.rb
```diff
@@ -36,4 +36,18 @@ class Search < ActiveRecord::Base; end
...
+    view_definition = "SELECT 'needle'::text AS haystack"
+    Search.connection.create_view :searches, sql_definition: view_definition
+    stream = StringIO.new
+
+    ActiveRecord::SchemaDumper.dump(Search.connection, stream)
+
+    output = stream.string
+
+    expect(output).to include "create_view :searches"
+    expect(output).not_to include "ar_internal_metadata"
```
calebthompson: Was schema_migrations covered by another spec? If so, can this be moved there so we don't duplicate the effort?

teoljungberg: It wasn't covered. So I added it.

calebthompson:
I think this may be the wrong approach. Moving conversation back to #194

calebthompson:
Merged in #198

PR #199, opened 2017-05-10
Travis: Extend matrix with Ruby 2.4

olleolleolle:
This PR updates the CI matrix adding a Ruby 2.4

It also updates the version of the Rubies already in place to the same versions in use at Rails.


derekprior:
Thanks!

PR #205, opened 2017-05-19
Use ActiveRecord::Migration.current_version instead of hardcoded value.

codebycliff:
If `ActiveRecord::Migration` responds to `current_version`, we should use that value instead of a hardcoded value. This pull request fixes that so that the generator is more future proof.


derekprior:
Thanks!

PR #206, opened 2017-05-21
Fix adapter config example to use correct method on config

pdebelak:
The correct method name is `database`, not `adapter`.


derekprior:
Thanks!

PR #208, opened 2017-06-07
Schema Dumper :: Sort Views by Name

icheishvili:
- updated schema dumper to sort view definitions by name so that schema generation is deterministic

We ran into an issue where running migrations in rails would cause large diffs to be generated as view definitions would shuffle around in the schema dump because different dev machines have different values of `pg_class.oid`, which Scenic::Adapters::Postgres::Views::#views_from_postgres orders by in its query.

This patch should make the dump process deterministic.

### DIFF
--- a/lib/scenic/schema_dumper.rb
+++ b/lib/scenic/schema_dumper.rb
@@ -22,7 +22,7 @@ module Scenic
     private
 
     def dumpable_views_in_database
-      @dumpable_views_in_database ||= Scenic.database.views.reject do |view|
+      @dumpable_views_in_database ||= Scenic.database.views.sort_by(&:name).reject do |view|
         ignored?(view.name)
       end
     end


calebthompson:
Thanks for your contribution! Because views can depend on other views, it's not as simple as sorting by name. All views need to be dumped after all tables _and views_ on which they depend need to be dumped first. Sorting by oid ensures this, at the cost of nondeterministic dumping. There's ongoing work on this in https://github.com/thoughtbot/scenic/pull/191.

icheishvili:
@calebthompson I read #191. Would something like this help in that case?

```WITH RECURSIVE view_deps AS (
  SELECT DISTINCT
    n.nspname AS namespace,
    c.relname AS viewname,
    c.relkind AS kind,
    c.oid,
    c2.relname AS viewdep
  FROM pg_rewrite AS r
  LEFT JOIN pg_class AS c
         ON r.ev_class = c.oid
        AND c.relkind IN ('m', 'v')
        AND c.relname NOT IN (SELECT extname FROM pg_extension)
  LEFT JOIN pg_depend AS d
         ON r.oid = d.objid
  LEFT JOIN pg_class AS c2
         ON d.refobjid = c2.oid
        AND c2.relkind IN ('m', 'v')
        AND c2.relname NOT IN (SELECT extname FROM pg_extension)
        AND c.oid != c2.oid
  LEFT JOIN pg_namespace AS n
         ON n.oid = c.relnamespace
  WHERE n.nspname = ANY (current_schemas(false))
), topological AS (
  SELECT 0 AS depth, vd1.viewname, vd1.namespace, vd1.kind, vd1.oid
  FROM view_deps vd1
  WHERE vd1.viewdep IS NULL
    AND NOT EXISTS (
      SELECT *
      FROM view_deps vd2
      WHERE vd2.viewname = vd1.viewname
      AND vd2.viewdep IS NOT NULL
    )
  UNION ALL
  SELECT topological.depth + 1 AS depth, vd1.viewname, vd1.namespace, vd1.kind, vd1.oid
  FROM view_deps vd1
  JOIN topological ON topological.viewname = vd1.viewdep
  WHERE vd1.viewdep IS NOT NULL
)
SELECT *
FROM topological
ORDER BY depth ASC, viewname ASC
;```

calebthompson:
It it helpful! @derekprior and I would certainly prefer a SQL solution to Ruby, but what it comes down to is that we need a sufficiently complex, real-world-ish example that we can test these solutions against and be confident that we don't break anything.

I'm sure you can understand that dumping an invalid schema is worse than dumping valid, but sometimes inconsistent, schemas.

oboxodo:
Has anyone else been working on this the last 3 years? Or maybe found some alternative workaround?

PR #212, opened 2017-06-17
Generate proper migration class name for non-public schemas

### LINKED ISSUE #187: Generation of view outside of public
When I generate a view that is not in public schema, it will be generated incorrectly.

Steps to reproduce:

1. Executing the command `create_view "schema.table_names", --materialized`, it generates a migration of the class `CreateSchematableNames` and a SQL file `schema_table_names_v01.sql`
2. When migration is executed, it will complain  uninitialized constant `CreateSchemaTableNames`
3. If you change to that class, it complains `No such file or directory @ rb_sysopen - /home/james/code/twitter-crawler/db/views/twitter.users_uks_v01.sql`

Found similar issue (https://github.com/thoughtbot/scenic/issues/95) and the resolution was merged into master but something may have changed the behaviour (https://github.com/thoughtbot/scenic/pull/124) (https://github.com/thoughtbot/scenic/issues/125)

Will do more investigation but the workaround was to rename the class and sql file.

pdebelak:
Previously, the code would take `"non_public.search"` and turn it into `NonPublicsearches` by calling `#camelize`, gsubbing out the period, and then calling `#pluralize`. This pull capitalizes letters after periods before gsubbing them out, so you end up with `NonPublicSearches` as expected.

Based on my understanding, this should fix #187, but I might be missing part of that issue because the person who opened it indicated they needed to rename the sql file, but I don't think that should be necessary based on how I see the code working.

### DIFF
--- a/lib/generators/scenic/view/view_generator.rb
+++ b/lib/generators/scenic/view/view_generator.rb
@@ -56,7 +56,7 @@ module Scenic
 
         def migration_class_name
           if creating_new_view?
-            "Create#{class_name.gsub('.', '').pluralize}"
+            "Create#{cleaned_class_name}"
           else
             "Update#{class_name.pluralize}ToVersion#{version}"
           end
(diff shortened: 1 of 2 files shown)


pdebelak:
Note: I can't seem to see the codeclimate issue (I get a 403) so I'm not sure what it is unhappy about.

magicalbanana:
Why was this never merged?

PR #214, opened 2017-07-14
Materialized Views - Add option "WITH NO DATA"

serg-kovalev:
This is amazing gem. Thank you for building and maintaining it!

At least for me, it happens so,  that I need to run migrations faster. Some huge materialized views might be refreshed for XX minutes/hours.
I'd propose to add an option `WITH NO DATA` for materialized views, to be able to call refresh of it later in some background logic.

What do you think? Is it a good candidate to be added into core functionality of the gem?

### DIFF
--- a/lib/generators/scenic/materializable.rb
+++ b/lib/generators/scenic/materializable.rb
@@ -10,21 +10,21 @@ module Scenic
           required: false,
           desc: "Makes the view materialized",
           default: false
-        class_option :materialized_no_data,
+        class_option :no_data,
           type: :boolean,
           required: false,
-          desc: "Makes the view materialized with NO DATA",
+          desc: "Adds WITH NO DATA when materialized view creates/updates",
           default: false
       end
 
       private
 
       def materialized?
-        options[:materialized] || options[:materialized_no_data]
+        options[:materialized]
       end
 
-      def materialized_no_data?
-        options[:materialized_no_data]
+      def no_data?
+        options[:no_data]
       end
     end
   end
--- a/lib/generators/scenic/materializable.rb
+++ b/lib/generators/scenic/materializable.rb
@@ -10,12 +10,21 @@ module Scenic
           required: false,
           desc: "Makes the view materialized",
           default: false
+        class_option :materialized_no_data,
+          type: :boolean,
+          required: false,
+          desc: "Makes the view materialized with NO DATA",
+          default: false
       end
 
       private
 
       def materialized?
-        options[:materialized]
+        options[:materialized] || options[:materialized_no_data]
+      end
+
+      def materialized_no_data?
+        options[:materialized_no_data]
       end
     end
   end
--- a/lib/generators/scenic/view/templates/db/migrate/update_view.erb
+++ b/lib/generators/scenic/view/templates/db/migrate/update_view.erb
@@ -4,7 +4,7 @@ class <%= migration_class_name %> < <%= activerecord_migration_class %>
     update_view <%= formatted_plural_name %>,
       version: <%= version %>,
       revert_to_version: <%= previous_version %>,
-      materialized: true
+      materialized: true<%= ", materialized_no_data: true" if materialized_no_data? %>
   <%- else -%>
     update_view <%= formatted_plural_name %>, version: <%= version %>, revert_to_version: <%= previous_version %>
   <%- end -%>
--- a/lib/generators/scenic/view/templates/db/migrate/update_view.erb
+++ b/lib/generators/scenic/view/templates/db/migrate/update_view.erb
@@ -4,7 +4,7 @@ class <%= migration_class_name %> < <%= activerecord_migration_class %>
     update_view <%= formatted_plural_name %>,
       version: <%= version %>,
       revert_to_version: <%= previous_version %>,
-      materialized: true<%= ", materialized_no_data: true" if materialized_no_data? %>
+      materialized: <%= no_data? ? "{ no_data: true }" : true %>
   <%- else -%>
     update_view <%= formatted_plural_name %>, version: <%= version %>, revert_to_version: <%= previous_version %>
   <%- end -%>
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -131,10 +131,13 @@ module Scenic
       #   in use does not support materialized views.
       #
       # @return [void]
-      def create_materialized_view(name, sql_definition, no_data = false)
+      def create_materialized_view(name, sql_definition, no_data: false)
         raise_unless_materialized_views_supported
-        no_data_option = no_data ? "\nWITH NO DATA" : ""
-        execute "CREATE MATERIALIZED VIEW #{quote_table_name(name)} AS #{sql_definition}#{no_data_option};"
+        execute <<-SQL
+  CREATE MATERIALIZED VIEW #{quote_table_name(name)} AS
+  #{sql_definition}
+  #{'WITH NO DATA' if no_data};
+        SQL
       end
 
       # Updates a materialized view in the database.
@@ -154,12 +157,12 @@ module Scenic
       #   in use does not support materialized views.
       #
       # @return [void]
-      def update_materialized_view(name, sql_definition, no_data = false)
+      def update_materialized_view(name, sql_definition, no_data: false)
         raise_unless_materialized_views_supported
 
         IndexReapplication.new(connection: connection).on(name) do
           drop_materialized_view(name)
-          create_materialized_view(name, sql_definition, no_data)
+          create_materialized_view(name, sql_definition, no_data: no_data)
         end
       end
 
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -122,6 +122,8 @@ module Scenic
       #
       # @param name The name of the materialized view to create
       # @param sql_definition The SQL schema that defines the materialized view.
+      # @param no_data [Boolean] Default: false. Set to true to create materialized view fast w/o
+      # feeding it with data. You will then need to `REFRESH MATERIALIZED VIEW`
       #
       # This is typically called in a migration via {Statements#create_view}.
       #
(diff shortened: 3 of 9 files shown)


lib/generators/scenic/materializable.rb
```diff
@@ -10,12 +10,21 @@ module Materializable
           required: false,
           desc: "Makes the view materialized",
           default: false
+        class_option :materialized_no_data,
```
derekprior: I think I'd prefer these to be separate options: `--materialized` and `--no-data`. I could see `--no-data` possibly implying `--materialized` but I'm not certain.

lib/scenic/adapters/postgres.rb
```diff
@@ -122,16 +122,19 @@ def drop_view(name)
...
+      # feeding it with data. You will then need to `REFRESH MATERIALIZED VIEW`
       #
       # This is typically called in a migration via {Statements#create_view}.
       #
       # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
       #   in use does not support materialized views.
       #
       # @return [void]
-      def create_materialized_view(name, sql_definition)
+      def create_materialized_view(name, sql_definition, no_data = false)
```
derekprior: Can we use a keyword argument here, please (`no_data: false`)?

lib/scenic/statements.rb
```diff
@@ -36,7 +36,7 @@ def create_view(name, version: nil, sql_definition: nil, materialized: false)
       sql_definition ||= definition(name, version)
 
       if materialized
-        Scenic.database.create_materialized_view(name, sql_definition)
+        Scenic.database.create_materialized_view(name, sql_definition, materialized_no_data)
```
derekprior: I'd also like to use a keyword argument here. I think it will make the boolean more understandable.

lib/scenic/statements.rb
```diff
@@ -82,7 +82,7 @@ def drop_view(name, revert_to_version: nil, materialized: false)
     # @example
     #   update_view :engagement_reports, version: 3, revert_to_version: 2
     #
-    def update_view(name, version: nil, sql_definition: nil, revert_to_version: nil, materialized: false)
+    def update_view(name, version: nil, sql_definition: nil, revert_to_version: nil, materialized: false, materialized_no_data: false)
```
derekprior: Can we use `no_data` instead?

lib/generators/scenic/view/templates/db/migrate/update_view.erb
```diff
@@ -4,7 +4,7 @@ class <%= migration_class_name %> < <%= activerecord_migration_class %>
     update_view <%= formatted_plural_name %>,
       version: <%= version %>,
       revert_to_version: <%= previous_version %>,
-      materialized: true
+      materialized: true<%= ", materialized_no_data: true" if materialized_no_data? %>
```
derekprior: To maintain the current style, this option should be rendered on its own line. Inserting the comma will be difficult there if we don't *always* render the `no_data` option, so I'd vote for just always rendering the option (true or false). This has the added advantage of making it more obvious the option exists.

lib/scenic/adapters/postgres.rb
```diff
@@ -122,16 +122,19 @@ def drop_view(name)
...
         raise_unless_materialized_views_supported
-        execute "CREATE MATERIALIZED VIEW #{quote_table_name(name)} AS #{sql_definition};"
+        no_data_option = no_data ? "\nWITH NO DATA" : ""
+        execute "CREATE MATERIALIZED VIEW #{quote_table_name(name)} AS #{sql_definition}#{no_data_option};"
```
derekprior: I think we could replace these two lines with something like:

```ruby
execute <<~SQL
  CREATE MATERIALIZED VIEW #{quote_table_name(name)} AS
  #{sql_definition}
  #{"WITH NO DATA" if no_data};
SQL
```

lib/scenic/adapters/postgres.rb
```diff
@@ -144,17 +147,19 @@ def create_materialized_view(name, sql_definition)
...
       # @param sql_definition The SQL schema for the updated view.
+      # @param no_data [Boolean] Default: false. Set to true to create materialized view fast w/o
+      # feeding it with data. You will then need to `REFRESH MATERIALIZED VIEW`
       #
       # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
       #   in use does not support materialized views.
       #
       # @return [void]
-      def update_materialized_view(name, sql_definition)
+      def update_materialized_view(name, sql_definition, no_data = false)
```
derekprior: Let's use a keyword argument here as well. I think it will improve readability at call sites.

lib/scenic/statements.rb
```diff
@@ -21,7 +21,7 @@ module Statements
     #     SELECT * FROM users WHERE users.active = 't'
     #   SQL
     #
-    def create_view(name, version: nil, sql_definition: nil, materialized: false)
+    def create_view(name, version: nil, sql_definition: nil, materialized: false, materialized_no_data: false)
```
derekprior: I think we should consider going the route many rails migration statements take with extensions to options. So in my migration, I'd be able to do something like:

```ruby
create_view :search_results, version: 1, materialized: { no_data: true }
```

That would imply both materialized and no data. A view would be materialized if the option was present. That might require some gymnastics for reading the option, but I think the use is much clearer in that case. What do you think? cc @calebthompson

lib/scenic/view.rb
```diff
@@ -22,15 +22,20 @@ class View
     # @return [Boolean]
     attr_reader :materialized
 
+    # True if the view is materialized with NO DATA
+    # @return [Boolean]
+    attr_reader :materialized_no_data
```
derekprior: just `no_data` is fine here, I think.

lib/scenic/view.rb
```diff
@@ -22,15 +22,20 @@ class View
...
+
     # Returns a new instance of View.
     #
     # @param name [String] The name of the view.
     # @param definition [String] The SQL for the query that defines the view.
     # @param materialized [String] `true` if the view is materialized.
-    def initialize(name:, definition:, materialized:)
+    def initialize(name:, definition:, materialized:, materialized_no_data: false)
```
derekprior: I think just `no_data` is fine here.

lib/scenic/view.rb
```diff
@@ -42,13 +47,14 @@ def ==(other)
 
     # @api private
     def to_schema
-      materialized_option = materialized ? "materialized: true, " : ""
+      materialized_option = materialized || materialized_no_data ? "materialized: true, " : ""
```
derekprior: Here I think we'd only want to be materialized if `materialized` was set. This is not a public interface really so I think we can be a bit more strict. If you want materialized and no data, you'd have to pass both. If you pass materialized: false and no_data: true, we can either ignore no_data or raise.

calebthompson: Postgres will probably "raise" for us in that case as well, which we may or may not wish to avoid.

lib/generators/scenic/view/templates/db/migrate/update_view.erb
derekprior: If we go this route:

```ruby
create_view :search_results, version: 1, materialized: { no_data: true }
```

Then we'd just set the materialized option to whatever the generator sets it to, which will either be a hash or false. We'd lose the discoverability bit, but I think that's ok.

lib/scenic/statements.rb
calebthompson: That's a reasonable interface, especially given the current API for migrations. I agree about `materialized_no_data` => `no_data` changes in general.

calebthompson:
> This is amazing gem. Thank you for building and maintaining it!

Thank you for the kind words, and for the pull request.

lib/scenic/view.rb
serg-kovalev: @derekprior @calebthompson Thank you very much for detailed review. I will try to fix everything you pointed out to me.

lib/generators/scenic/materializable.rb
serg-kovalev: @derekprior Yeah, you're correct. So, let's have separate options

serg-kovalev:
@derekprior @calebthompson 
I know you're busy. I'd love to hear your feedback.

Also I want to share with you some of my findings. I faced with an issue when it's impossible to concurrently refresh materialized view, when it's not yet been populated (say, you run migration with `no_data` option and never did `REFRESH MATERIALIZED VIEW`. So, it's kind of expected behavior, but I would like to share with you the hack I use to solve this issue:

```ruby

module MViewRefreshable
  extend ActiveSupport::Concern

  private

  def concurrently_refreshable?
    begin
      concurrently_refreshable = true
      self.first
    rescue ActiveRecord::StatementInvalid => e
      raise e unless e.message =~ /Use the REFRESH MATERIALIZED VIEW/i
      concurrently_refreshable = false
    end
    concurrently_refreshable
  end
end
```

lib/scenic/view.rb
```diff
@@ -22,15 +22,20 @@ class View
...
+    # @return [Boolean]
+    attr_reader :no_data
+
     # Returns a new instance of View.
     #
     # @param name [String] The name of the view.
     # @param definition [String] The SQL for the query that defines the view.
     # @param materialized [String] `true` if the view is materialized.
-    def initialize(name:, definition:, materialized:)
+    def initialize(name:, definition:, materialized:, no_data: false)
```
derekprior: I'm not sure `Scenic::View` needs to do anything with `no_data`. As best I can tell, we only use this construct when dumping the schema. At that point, I don't believe we know if the view was created with no data or not? Indeed, the only call site for `View.new` does not add the `no_data` parameter at all.

This effectively means that views created by loading schema would never use `WITH NO DATA`. It would be a migration-time only option.

This makes me a little nervous because theoretically that could allow production and development inconsistencies. For example, if in my migration I do `create_view :searches, materialized: { no_data: true }` and run that in development and production I'd be left with an empty `searches` view. If in development I later run `rake db:schema:load` then my view will now be populated in development only which may lead to some confusion.

Should we be concerned about that? Thoughts, @calebthompson @serg-kovalev ?

derekprior:
Opened #241 with some style cleanup and the removal of the changes to `view.rb`. Thanks for the PR and sorry for the delay. Assuming tests pass, we'll get this in and released.

serg-kovalev:
@derekprior I'm sorry for being silent for a while. I was too busy with my day job. Thanks for merging it!

PR #215, opened 2017-08-18
Make #readonly? method public in example

xymbol:
According to `rails`, the `#readonly?` method should be public. https://github.com/rails/rails/blob/a05845e0a699e169c11962212200dc8e72934dff/activerecord/lib/active_record/core.rb#L483


derekprior:
Thanks!

PR #223, opened 2017-11-07
Update README about dropping materialized views

devonestes:
An issue was opened up about how to do this, and indeed the README was
a little lacking in guidance for how to drop a materialized view. I
think this should cover it by offering an example.


calebthompson:
Thanks @devonestes

PR #233, opened 2018-01-21
Fix mysql adapter link in README.md.

simi:


calebthompson:
Thanks @simi and @stevensona!

PR #237, opened 2018-02-14
Support multiple views paths

smdern:
I needed to use this in a rails engine and didn't want to have to copy db_views over for every rails app. 

This makes it so the the definition is searched for in multiple paths when trying to call `to_sql`. Generator paths aren't affected. I leveraged Rails' `config.paths` for configuring additional `db/views` paths. I could change it so that the paths are apart of `Scenic::Configuration`, I just thought putting it in `Rails.config.paths['db/views']` would pair nicely with `Rails.config.paths['db/migrations']` works.

I'm able to make this work with a rails engine by using the following in `engine.rb`

```ruby
    initializer :append_migrations do |app|
      app.config.paths["db/views"] << root.join("db/views")

      unless app.root.to_s.match? root.to_s
        config.paths["db/migrate"].expanded.each do |expanded_path|
          app.config.paths["db/migrate"] << expanded_path
        end
      end
    end
```

### DIFF
--- a/lib/scenic/definition.rb
+++ b/lib/scenic/definition.rb
@@ -15,12 +15,12 @@ module Scenic
     end
 
     def find_definition
-      definition = views_paths.expanded.flat_map do |directory|
+      definition = views_paths.flat_map do |directory|
         Dir.glob("#{directory}/**/#{filename}")
       end.first
 
       unless definition
-        raise "Unable to locate #{filename} in #{views_paths.expanded}"
+        raise "Unable to locate #{filename} in #{views_paths}"
       end
 
       definition
@@ -41,7 +41,7 @@ module Scenic
     private
 
     def views_paths
-      Rails.application.config.paths["db/views"]
+      Rails.application.config.paths["db/views"].expanded
     end
 
     def filename
--- a/lib/scenic/definition.rb
+++ b/lib/scenic/definition.rb
@@ -7,11 +7,23 @@ module Scenic
     end
 
     def to_sql
-      File.read(full_path).tap do |content|
+      File.read(find_definition).tap do |content|
         if content.empty?
           raise "Define view query in #{path} before migrating."
         end
       end
+    end
+
+    def find_definition
+      definition = views_paths.expanded.flat_map do |directory|
+        Dir.glob("#{directory}/**/#{filename}")
+      end.first
+
+      unless definition
+        raise "Unable to locate #{filename} in #{views_paths.expanded}"
+      end
+
+      definition
     end
 
     def full_path
@@ -27,6 +39,10 @@ module Scenic
     end
 
     private
+
+    def views_paths
+      Rails.application.config.paths["db/views"]
+    end
 
     def filename
       "#{@name}_v#{version}.sql"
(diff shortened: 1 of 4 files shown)


coding-chimp:
I pretty much have the same use-case. Would live to see this get merged! 👍

MarcReniu:
Hello!! Anyone know when this PR will be merged? Thanks

lib/scenic/definition.rb
```diff
@@ -7,13 +7,25 @@ def initialize(name, version)
...
     def to_sql
-      File.read(full_path).tap do |content|
+      File.read(find_definition).tap do |content|
         if content.empty?
           raise "Define view query in #{path} before migrating."
         end
       end
     end
 
+    def find_definition
```
derekprior: Can you rename this `definition_path`?

lib/scenic/definition.rb
```diff
@@ -7,13 +7,25 @@ def initialize(name, version)
...
+      definition = views_paths.flat_map do |directory|
```
derekprior: Should this be `flatten.find {}` instead of `flat_map {}.first`?

spec/scenic/definition_spec.rb
```diff
@@ -2,22 +2,51 @@
 
 module Scenic
   describe Definition do
+    describe "find_definition" do
```
derekprior: Should this be a tested public method? Or should it be private and tested as part of other tests in the app?

smdern:
> Further, are there methods on `Definition` that we no longer need? (`path`, `full_path`)?

These could probably go away. Currently they're being used in a helper for creating new views using the generator, so maybe they could instead be changed to `target_path` or something. It looks like `path` could at least go away since it's just being used by `full_path`

vladimirtemnikov:
Hi, really waiting for this 👍

tomasc:
Hi, would be great to have this merged! Just ran into this when trying to use scenic from engines.

derekprior:
Rails 6 adds native support for multiple databases. At this point, I’m inclined to see how we would be able to work with that.

tomasc:
@derekprior an alternative approach would be to make the `full_path` used in [definition.rb](https://github.com/scenic-views/scenic/blob/master/lib/scenic/definition.rb#L18) configurable – by way of being able to pass `root_path` in the configuration:

```ruby
Scenic.configure do |config|
  config.root_path = MyEngine.root
end
```

In any case, the current setup prevents the gem from being used from within Rails engines as it looks up the `db/views` under the engine's test app. This is not related to multi-database support in Rails 6 I believe.

Would you be interested in a PR implementing the above?

halexarmenta:
I agree with @tomasc , it would be nice to have a configuration called **root_path** like the one that @tomasc propose or **views_path** to be more specific to set the views path, i thinks this is a better approach than the one on this PR because it is more flexible than overriding a definition.

@derekprior Any provisional solution on how to use scenic inside an engine without copying files?

derekprior:
Long time no review on this PR. We've had a number of PRs for this feature (see https://github.com/scenic-views/scenic/labels/view-paths) so I think it merits further consideration. I think this is the closest to something we could accept as it supports multiple search paths. I hope to look at this closer in the next couple of weeks.

PR #232, opened 2018-01-15
Add Ruby 2.5 to travis build matrix

devonestes:
Maybe it's time to drop support for those versions of Ruby that are no
longer being updated? 2.1 is currently end of life, and 2.2 is going to
be end of life'd at the end of March.

PR #257, opened 2018-10-29
Add Rails 5.2 build

smudge:
👋 Rails 5.2 was released in April. I've updated your Travis build to match all of the versions currently included in Rails' [Maintenance Policy](https://guides.rubyonrails.org/maintenance_policy.html).

I also followed-up on the version support conversation in #232 and:
- Removed unsupported versions from the `Appraisals` file (+re-generated all of the gemfiles off of this)
- Updated `required_ruby_version` in the gemspec.


spec/dummy/app/models/application_record.rb
```diff
@@ -0,0 +1,5 @@
+if Rails::VERSION::STRING >= '5.2.0'
+  class ApplicationRecord < ActiveRecord::Base
```
smudge: Models generated on Rails 5.2 (and edge) were expecting an `ApplicationRecord` to have been defined in the dummy app.

PR #264, opened 2018-12-21
Add the OracleEnhanced adapter to the README list

fractaledmind:


derekprior:
Thanks @fractaledmind. I looked through the adapter really quickly and noticed there's definitely some features not supported (index reapplication, for example). I think those would be worth calling out in your README.

PR #268, opened 2019-02-03
Update hound and rubocop configuration

derekprior:
When we moved to our own organization from the thoughtbot org, we lost
the hound-provided rubocop config associated with the thoughtbot org.
This change updates to the prefered hound config format that references
our own `.rubocop.yml` file.

How did I settle on the rubocop rules? Well, I ran rubocop on the
codebase and either fixed what came up or added exceptions. We have lots
of line length violations that I didn't fix. For the most part, I like
the encouragement to stay at 80 characters, but sometimes it's not worth
it. I think getting the hound reminder in the PR is sufficient enough
for us to have a conversation.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,129 @@
+AllCops:
+  TargetRubyVersion: 2.3.0
+  Exclude:
+    - "tmp/**/*"
+    - "bin/*"
+    - "spec/dummy/**/*"
+
+Bundler/OrderedGems:
+  Enabled: false
+
+Gemspec/OrderedDependencies:
+  Enabled: false
+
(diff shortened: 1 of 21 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,129 @@
...
+    find_all: select
+Style/ConditionalAssignment:
+  Enabled: false
+Style/ClassAndModuleChildren:
+  Enabled: true
+  Exclude:
+    - "spec/**/*"
+Style/Documentation:
+  Enabled: false
+Style/FrozenStringLiteralComment:
```
derekprior: We should probably revisit this one in a future change. Just didn't want to do it right now.

PR #280, opened 2019-08-20
Allow override of view definition filenames

bhaibel:
I'm working on a project where we're using Postgres multi-schema with views, which would lead to views that would theoretically be created with a call like this one:

```
create_view 'schema_name.view_name'
```

However, because Scenic requires the view name to be the same as the definition file's name, the dot in the view name when specifying the target schema means that I can't really use schema definition files.

This PR adds a `defined_as` flag to allow users to override the definition file name.

### DIFF
--- a/lib/scenic/statements.rb
+++ b/lib/scenic/statements.rb
@@ -9,6 +9,10 @@ module Scenic
     # @param sql_definition [String] The SQL query for the view schema. An error
     #   will be raised if `sql_definition` and `version` are both set,
     #   as they are mutually exclusive.
+    # @param defined_in [String] The file where the view is defined. If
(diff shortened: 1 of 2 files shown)


bhaibel:
I'm noticing a lot of lint comments that I'm at a loss for fixing in ways that would actually make the code better -- e.g. things like "unused argument" for arguments that seem to be mirrored for reversibility. How would you like me to address those?

gumchum:
I actually ran into this exact same issue. https://github.com/scenic-views/scenic/pull/282 is my own attempt at fixing this (and another small bug related to this problem).

derekprior:
I believe 282 addresses this use case and has now been merged.

PR #284, opened 2019-10-29
Allow dropped views to be cascaded

dansteele:
Also added generated dummy files to the `.gitignore`


dansteele:
Ah, didn't think about rolling back which obviously this isn't compatible with.

PR #289, opened 2020-02-02
Make `create_view` with `version` reversible

derekprior:
Prior to this change, `create_view :greetings, version: 2` would error
because it was reversed to `drop_view :greetings, version: 2`. The
`drop_view` method on the Postgres adapter only takes a view name, so we
shouldn't record additional arguments when inverting.


calebthompson:
👍

PR #290, opened 2020-02-02
Update build matrix

derekprior:
Our existing build matrix was a little out of date. We were missing
Rails 6 and Ruby 2.7. I could have just added those and called it a day,
but our build matrix was already getting unwieldy.

I started by removing Ruby 2.3, which is no longer receiving updates. I
then removed Rails versions older than 5.2 for the same reason. Finally,
I removed the Rubies between 2.4 and 2.7 on the grounds that testing on
the oldest currently-supported ruby and the newest is sufficient.

Scenic's Rails and Ruby version support remains unchanged, but we're
limiting CI to a reasonable combination based on ongoing support from
Ruby and Rails. We're saving build time and CO2 emissions.


calebthompson:
🌳

PR #293, opened 2020-02-09
Use GitHub Actions for CI

derekprior:
Travis hasn't been my preferred CI provider for quite some time, but the
ease with which you can set up a build matrix meant it still made sense
to use for open source libraries. I've been curious about GitHub Actions
for quite some time and after kicking the tires on them a bit at work, I
wanted to dive in deeper. This change moves our entire test suite to
Actions. Beyond personal taste (and the fact that I work at GitHub), why
would I do this?

1. Speed: The jobs kick off almost immediately with no wait time and
more concurrency than Travis. They also execute very quickly. We
recently trimmed our build matrix back significantly, but Travis builds
would still take at least 8 minutes. Builds with Actions complete in 3
minutes.
2. General Purpose: Setting up CI with actions was a bit more involved
than Travis, but workflow files are much more flexible. We can start to
automate other things (publishing releases, etc) with the same
knowledge.
3. UI: We can see the results of each build directly in the pull
request UI without a click through.

The most interesting "new" step in the workflow file is probably the
explicit lock file generation. We use the contents of the lock file as
the cache key for dependency caching. Since we don't check in any lock
files, we need to generate one without installing dependencies so we can
see if the cache is valid and should be restored. This explicitness in
how caching works highlighted to me that caching probably wasn't working
on Travis at all for similar reasons.

Beyond replicating `.travis.yml` as a Workflow, I also made some changes
to the test setup itself. These weren't strictly necessary but since I
was thinking about how all the tests work anyway, I decided to take the
plunge and remove Appraisal. There's nothing that needs to change
between our builds other than the version of Rails and we can get the
functionality we need for that with a few lines of Ruby in the Gemfile.
This made test setup simpler. We also don't have to create new Gemfiles
for every version of Rails we want to test with.


derekprior:
This is coming along okay. When I remove the cache step, the tests run and I get the same results that I do on Travis but faster. I have a couple of things left to figure out:

1. Why is the cache not working?
2. Once it works for bundler, is it possible to also cache the `libpq-dev` dependency? Installing that takes a bit.
3. Is there anything akin to `allowed_failures` from Travis? We want to test against rails master, but we don't want a failed check status if it fails. It is currently failing and we should probably look into that, but it's not something that should block work.

derekprior:
This is in a state where I think it's now good to go and I've squashed the 35+ commits I made as I learned how this all works. I've updated the PR comment as well. We have allowed failures on Rails edge and Gemfile caching (which I don't think was working on Travis). 

From what I've seen, the test suite kicks off much quicker, has more parallelism, and completes quicker (8 minutes -> 3 minutes). 🎉

derekprior:
I've had some folks with some more actions experience look this over and I was able to simplify the workflow a bit based on their feedback. I'm going to go ahead and merge this.

PR #294, opened 2020-02-20
Adds language fallback for example

henrytseng:
Signed-off-by: Henry Tseng <info@heycanvas.com>


derekprior:
What problem does this solve?

PR #298, opened 2020-04-28
Implement custom views path

DNA:
Hi there!

Here at work we use two database adapters, SQL Server and SQLite (for quick tests before CI/CD). Since some syntax between those two are incompatible, we need to have two SQL files, and those can't live in the same directory.

The best solution we found was to add a configuration attribute to define the views path directory, so we can edit it inside Rails like that:

```ruby
adapter = Rails.env.test? ? 'sqlite' : 'sqlserver'

Scenic.configure do |config|
  config.views_path = "db/views/#{adapter}"
end
```

Since more people could benefit from this change, I've decided to open a PR.
Let me know if it's is OK or if I missed something 😃

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ changelog, see the [commits] for each version via the version links.
 
 [commits]: https://github.com/scenic-views/scenic/commits/master
 
+## [Unreleased]
+
+Added
+
+* Add a configuration value for changing the default path for the views files
+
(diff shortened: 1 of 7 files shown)


derekprior:
We have a number of proposed changes (see: https://github.com/scenic-views/scenic/labels/view-paths) that cover this same functionality. I think https://github.com/scenic-views/scenic/pull/237 is closest to something we would consider merging in that covers multiple use cases. Swapping one static path for another would solve some problems but create others.

PR #300, opened 2020-05-07
Add security policy

derekprior:
- Update Caleb's last name
- Add Security Policy


calebthompson:
Thanks for updating me!

PR #302, opened 2020-05-13
Update code snippets in README

claasz:
Since Rails 5.0, all models inherit from `ApplicationRecord`, not `ActiveRecord::Base`


calebthompson:
Great, thanks!

Issue #242, opened 2018-05-26
Views with regexp and postgres


lucasmpb:
I think there is an issue, when views with regular expressions are not correctly been including on schema.rb file.
As an example, if we have a view like
```
SELECT * FROM table WHERE name::text ~ 'qe(\d+)'::text
```
On the schema.rb file is added as
```
SELECT * FROM table WHERE name::text ~ 'qe(d+)'::text
```
(note the missing `\`) breaking the regexp

derekprior:
I looked into this, and the issue seemingly exists either in PG or in `pg_get_viewdef` itself. I created a test case for this and did some debugging. We run the following query:

```SQL
            SELECT
              c.relname as viewname,
              pg_get_viewdef(c.oid) AS definition,
              c.relkind AS kind,
              n.nspname AS namespace
            FROM pg_class c
              LEFT JOIN pg_namespace n ON n.oid = c.relnamespace
            WHERE
              c.relkind IN ('m', 'v')
              AND c.relname NOT IN (SELECT extname FROM pg_extension)
              AND n.nspname = ANY (current_schemas(false))
            ORDER BY c.oid
          SQL
```

When we execute this query we get back a `PG::Result`. This PG::Result strips the `\`.

I don't think there's anything we can do here. I'm going to close this, but if you think I'm mistaken, please ping the issue again.

derekprior:
You can try it out for yourself debugging this branch, if you'd like: 
https://github.com/scenic-views/scenic/compare/dp-regex-views?expand=1

bensheldon:
I'm having this problem with regexp, but I'm seeing different behavior depending on whether the views are being created as part of a migration, or as part of `schema:load`. 

- Created as part of migration: the resulting definition in `pg_mat_views` correctly has a backslash. Also, running `pg_get_viewdef('my_view')` correctly returns a definition with a backslash too.
- Created as part of `bin/rails db:schema:load` (or `db:reset`): the resulting definition in pg_mat_views` incorrectly does not have a backslash, nor does `pg_get_viewdef('my_view')`

I think the problem is that the backslash requires a double-backslash when it's being outputted into the `schema.rb` file in the HEREDOC. 

I'll look further at the debugging branch.

calebthompson:
> I'm seeing different behavior depending on whether the views are being created as part of a migration, or as part of schema:load.

[...]

PR #303, opened 2020-06-02
Fix serialization of views that contain regexp backslashes

bensheldon:
Based on #242, Views that contain backslashes are not properly serialized into the schema. I think there may be a more elegant way to do this than gsubbing backslashes, but the underlying issue is that a string is interpolated directly into a Ruby HEREDOC which does not appropriately escape backslashes.

I had trouble passing the test assertion that verifies that the schema dump contains the view_definition because of what I think is whitespace inconsistencies.

### DIFF
--- a/lib/scenic/view.rb
+++ b/lib/scenic/view.rb
@@ -46,7 +46,7 @@ module Scenic
 
       <<-DEFINITION
   create_view #{name.inspect}, #{materialized_option}sql_definition: <<-\SQL
-    #{definition.gsub("\\", "\\\\\\").indent(2)}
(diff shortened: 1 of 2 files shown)


bensheldon:
Here is a simpler representation of the backslashing necessary:

```ruby
string = "\\backslash"
code = <<-HEREDOC
<<-DOC
  #{string.gsub('\\', '\\\\\\')}
DOC
HEREDOC

result = eval(code).strip

puts result
puts result == string.strip
```

derekprior:
Superseded by #346 -- better late than never, eh?

Issue #261, opened 2018-12-06
Use scenic within a rails engine / component


nathanpalmer:
We would like to use scenic from within a rails engine. However, the path to the view is not coming up correctly. In my example I have this.

```
- app
- db
- components
    - products
    - db
      - views
```

The `products` is the rails engine that is referenced by the main app by this.

```ruby
path 'components' do
  gem 'products'
end
```

and to get rails to recognize regular migrations this is put in the `engine.rb`

```ruby
module Products
  class Engine < ::Rails::Engine
    isolate_namespace Pipedrive

    initializer :append_migrations do |app|
      unless app.root.to_s.match root.to_s+File::SEPARATOR
        app.config.paths["db/migrate"].concat config.paths["db/migrate"].expanded
      end
    end

    config.generators do |g|
    end
  end
end
```

# Failing in the component directory

From inside the `components/products` directory when I run a `rake db:migrate` it seems to be looking for the view in `test/dummy/db/views` instead of `db/views`.

<details>
   <summary>20181206151701 CreateExpandedDeals: migrating</summary>

```
== 20181206151701 CreateExpandedDeals: migrating ==============================
-- create_view(:expanded_deals)
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

No such file or directory @ rb_sysopen - /Users/nathan/Source/test_app/components/products/test/dummy/db/views/expanded_deals_v01.sql
/Users/nathan/.rvm/gems/ruby-2.3.4/gems/scenic-1.4.1/lib/scenic/definition.rb:10:in `read'
/Users/nathan/.rvm/gems/ruby-2.3.4/gems/scenic-1.4.1/lib/scenic/definition.rb:10:in `to_sql'
/Users/nathan/.rvm/gems/ruby-2.3.4/gems/scenic-1.4.1/lib/scenic/statements.rb:142:in `definition'
/Users/nathan/.rvm/gems/ruby-2.3.4/gems/scenic-1.4.1/lib/scenic/statements.rb:36:in `create_view'
/Users/nathan/.rvm/gems/ruby-2.3.4/gems/activerecord-5.1.6.1/lib/active_record/migration.rb:849:in `block in method_missing'
/Users/nathan/.rvm/gems/ruby-2.3.4/gems/activerecord-5.1.6.1/lib/active_record/migration.rb:818:in `block in say_with_time'
[...]

PR #306, opened 2020-09-16
Configurable root path

tomasc:
This PR allows for configuration of the root path, where Scenic looks up its views:

```ruby
Scenic.configure do |config|
  config.root_path = MyEngine.root
end
```

The default value stays `Rails.root`.

This is a follow up to https://github.com/scenic-views/scenic/pull/237#issuecomment-513880655 and addresses #261.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,16 @@ meet your needs:
 Please note that the maintainers of Scenic make no assertions about the
 quality or security of the above adapters.
 
+**Can I use Scenic in Rails engines?**
+
+Yes, just update the path where Scenic looks up its views in its configuration:
+
+```ruby
+Scenic.configure do |config|
+  config.root_path = MyEngine.root
+end
+```
+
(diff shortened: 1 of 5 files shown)


derekprior:
@tomasc how would this work in cases where both the engine and the app itself want to use Scenic?

trevorrjohn:
I would think you need two config files for that, one in the engine and one at the app level. Or you could add a env var or argument to override the "default" path.

derekprior:
The configuration is currently a singleton, so I don't think that would work

derekprior:
We have a number of proposed changes (see: https://github.com/scenic-views/scenic/labels/view-paths) that cover this same functionality. I think https://github.com/scenic-views/scenic/pull/237 is closest to something we would consider merging in that covers multiple use cases. Swapping one static path for another would solve some problems but create others.

PR #313, opened 2020-12-09
Fix loading error on `Bundler.require`

pcreux:
Calling `Bundler.require` or `require 'scenic'` was throwing errors:
* `undefined method 'delegate' for Scenic::Adapters::Postgres::Indexes:Class`
* `uninitialized constant Scenic::SchemaDumper::ActiveRecord`

These errors don't happen when running rails commands since
`config/application.rb` requires 'rails' and 'active_record'
before calling `Bundler.require`.

It however prevents standalone scripts and executable (such as `sidekiqswarm`)
from calling `Bundler.require` before initializing the rails app.

Adding `Bundler.require` to `spec_helper.rb` reproduces this issue.

### DIFF
--- a/lib/scenic.rb
+++ b/lib/scenic.rb
@@ -1,3 +1,6 @@
+require "rails"
+require "active_record"
+
 require "scenic/configuration"
 require "scenic/adapters/postgres"
 require "scenic/command_recorder"
--- a/lib/scenic/schema_dumper.rb
+++ b/lib/scenic/schema_dumper.rb
@@ -1,5 +1,3 @@
-require "rails"
-
 module Scenic
   # @api private
   module SchemaDumper
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,3 +1,5 @@
+Bundler.require
+
 ENV["RAILS_ENV"] = "test"
 require "database_cleaner"
 


pcreux:
Oh no! CI is broken.

Screen Shot 2020-12-08 at 10 02 06 PM

krschacht:
@pcreux I have this issue as well. How have you worked around it? Did you figure out a way to use sidekiqswarm?

pcreux:
@krschacht Our `Gemfile` points to this branch. 🙄 

```ruby
gem 'scenic', github: 'retailzipline/scenic', branch: 'require-active-record'
```

@derekprior Looks like @krschacht is running into this issue as well. Could you have another look at this fix, please?

krschacht:
@derekprior The author of Sidekiq explained the problem with scenic. His note was, "Yeah, [scenic] is trying to extend Rails functionality at require time rather than in a Railtie initializer. Bad idea."

derekprior:
Hmmm, that's certainly not our intent. We try to load our monkey patches in `Scenic.load` which we call from our Railtie:

```
  class Railtie < Rails::Railtie
    initializer "scenic.load" do
      ActiveSupport.on_load :active_record do
        Scenic.load
      end
    end
  end
```

We probably need to do more inside the initializer, just not sure what.

krschacht:
For anyone else stuck on this issue, I ultimately solved it by removing the Scenic and adding this simple module:

```
module MigrationHelpers
    def create_view(name, opts = {})
        version = (opts != {}) ? opts[:version].to_i : 1
        path = Rails.root.to_s
        views_path = (Rails.root.to_s =~ /gems/) ? Rails.root.join("..", "..", "db", "views").to_s : Rails.root.join("db", "views").to_s

        sql = File.read("#{views_path}/#{name}_v#{"%02d" % version}.sql")

        execute("CREATE VIEW public.#{name} AS #{sql}")
    end

    def drop_view(name)
        execute("DROP VIEW #{name};")
    end
end
```

In all my migrations that need to `create_view` and `drop_view` I simply add `include MigrationHelpers` at the top of that migration file. This allowed all of my old scenic migrations and view files to keep working.

markedmondson:
Hey — we've been carrying a fork with this patch since 2020. Looks like this was inadvertently fixed in v1.9.0 by 736dee8f which removed the `unless ActiveRecord::SchemaDumper.private_instance_methods(false).include?(:ignored?)` block from `schema_dumper.rb`. That was the line that blew up during `Bundler.require` before ActiveRecord was loaded.

We've verified `Bundler.require(:default)` works cleanly on 1.9.0 without the explicit `require "active_record"` patch. Dropping our fork now :slightly_smiling_face:

PR #319, opened 2021-01-20
docker

gagalago:
add configuration to be able to run tests of scenic inside docker (doesn't prevent to continue to work as usual when everything is installed locally).

That allows to easily create and clean up the local setup needed to work on this gem with docker.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,8 @@ agree to abide by our [code of conduct].
 6. CI will run the test suite on all configured versions of Ruby and Rails.
    Address any failures.
 
+you can also run it (step 2 and 3) localy with docker: `docker-compose run scenic rake`.
+
 [good commit message]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
 
 Others will give constructive feedback.  This is a time for discussion and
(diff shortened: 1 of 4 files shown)


calebthompson:
I’m not sure that we want this as the setup runs fine without needing the additional dependency on docker.

If it’s easier for you to run the tests in docker, in my opinion I’d rather you keep this locally and not commit it.

gagalago:
> I’m not sure that we want this as the setup runs fine without needing the additional dependency on docker.
> 
> If it’s easier for you to run the tests in docker, in my opinion, I’d rather you keep this locally and not commit it.

From my point of view, it's not a dependency but more another way of working that can bring more contributors to this project without impacting actual contributors that can still continue to work as usual without docker.

I added docker to be able to easily create and clean my local setup to work on this gem.

It's perfectly fine for me to keep this only for me, but as I did it I just propose it, and then it's to the community to choose to integrate it or not.

derekprior:
I agree with @calebhearth, so I'm going to close this. One thing I'd be interested in is making sure we run pretty simply in codespaces

derekprior:
@gagalago see #345 for a similar implementation aimed specifically at codespaces but likely usable locally

Issue #325, opened 2021-03-23
Schema search path should not be included in schema.rb when a migration is executed.


rapito:
When executing a migration  **the schema search path** should not be included on the schema.rb since this is Environment dependant and **will break ** on a different configuration:

Example of schema.rb generated because of scenic:

```
  create_view "my_schema_search_path.foo", materialized: true, sql_definition: <<-SQL
      SELECT * from foo;
  SQL
  add_index "my_schema_search_path.foo", ["bar"], name: "index_on_foo_bar"
```

Expected schema.rb appendix is:

```
  create_view "foo", materialized: true, sql_definition: <<-SQL
      SELECT * from foo;
  SQL
  add_index "foo", ["bar"], name: "index_on_foo_bar"
```

derekprior:
Is it not reasonable to expect schema names to be consistent across environemnts?

rapito:
There are several use cases where the schema_search_path is what changes across the environments and the db stays the same, thats why activerecord does not include the search_path on the schema.rb.


> Is it not reasonable to expect schema names to be consistent across environments?

jdejong:
I agree with this approach since it directly aligns with ActiveRecord's approach. 

Additionally, many DBaaS providers use Schema or Database names as a way to uniquely identify databases and companies in their platform so you might not have control over either across your environments.

derekprior:
This seems like it would be a breaking change given that the reason this exists in the first place was due to users taking the opposite approach. However, given that Rails now has native multiple database support, it may be worth revisiting. I haven't done anything personally with multi db support though.

@calebthompson what do you think?

calebthompson:
> many DBaaS providers use Schema or Database names as a way to uniquely
identify databases and companies

I haven’t run into this before. Can you share a docs example for using
schema names here? Heroku does database names for hobby-tier databases
but schemas are new to me.

On April 15, 2021, GitHub ***@***.***> wrote:
> This seems like it would be a breaking change given that the reason
> this exists in the first place was due to users taking the opposite
> approach. However, given that Rails now has native multiple database
> support, it may be worth revisiting. I haven't done anything
> personally with multi db support though.
>
> @calebthompson <https://github.com/calebthompson> what do you think?
>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/scenic-views/scenic/issues/325#issuecomment-
820681280>, or unsubscribe
<https://github.com/notifications/unsubscribe-
[...]

PR #152, opened 2016-05-13
Adds view's namespace to the SchemaDumper

Willianvdv:
This adds the view's namespace when dumping a view. This is useful when views live in another schema than `public`. Before this change the dumper would dump `thoughtbot.scenic` as `scenic`. In our case this would collide with our `scenic` view that's living in the `public` namespace.

PR #327, opened 2021-04-08
Use view without namespace when dumping schema.

rapito:
## Why?

Without this, default behavior of `rake db:schema:load` is broken becuase Activerecord config option `schema search path` for Postgres adapters is automatically used to identify where is the schema being loaded to; By explicitly adding the namespace/search_path in the schema.rb it's wrong as you can run rake db:migrate from multiple computers and configurations which result in namespace_from_pc_1.view_name and namespace_from_pc_2.view_name on different commits.

## Changes

+ Added a new private method unscoped_view to make sure to not include the namespace when dumping the schema name
+ Updated tests to make sure the namespace is excluded rather than included.

## Extra

An additional justification is that schema.rb does not include namespaces ever, and the default Rails/Activerecord behavior should be conserved.

Fixes:  https://github.com/scenic-views/scenic/issues/325

### DIFF
--- a/lib/scenic/schema_dumper.rb
+++ b/lib/scenic/schema_dumper.rb
@@ -15,7 +15,7 @@ module Scenic
 
       dumpable_views_in_database.each do |view|
         stream.puts(view.to_schema)
-        indexes(view.name, stream)
+        indexes(view.unscoped_name, stream)
       end
     end
 
--- a/lib/scenic/view.rb
+++ b/lib/scenic/view.rb
@@ -41,11 +41,15 @@ module Scenic
     end
 
     # @api private
+    def unscoped_name
+      name.split('.').last
+    end
+
+    # @api private
     def to_schema
       materialized_option = materialized ? "materialized: true, " : ""
-
       <<-DEFINITION
-  create_view #{name.inspect}, #{materialized_option}sql_definition: <<-\SQL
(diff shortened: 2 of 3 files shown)


derekprior:
@rapito This is in direct conflict with https://github.com/scenic-views/scenic/pull/152, which introduced this behavior. Honestly, I have zero intention of ever using Scenic with multiple namespaces so I'm not sure what the correct solution is here. Can you convince me and/or @calebhearth which decision is correct?

I see your point that we can allow the schema search path to determine this for us. However, what about the case where your app actually uses multiple schemas? If we rely on schema search path alone, wouldn't all views be created in whatever the first listed schema is when in reality we might want some in the first listed schema and some others in the second?

How does Rails itself handle this with tables?

cc @Willianvdv

rapito:
To that end I would suggest making the scenic migration accept optional parameters to include the schema search path. 

And to still allow specifying schemas, the method should not include the schema on the scenic view name but as an argument of the method.
I.e: 
```rb
create_view "tablename", schema: "schema_name"
```

The gem would translate that to "schema_name.tablename" when executing it.

Furthermore, the schema parameter should only be set if the gem config option `include_search_path_on_schema` is turned on. 

Imo, Ideally the primary feature that the gem should support is creating the scenic views. Everything else should be secondary such as supporting multiple schemas, and if they break default behaviors or patterns then they should be optional.

Issue #253, opened 2018-10-02
Error on Migration: wrong number of arguments (given 2, expected 1)/usr/local/bundle/gems/highline-1.7.10/lib/highline.rb:620:in `say'


Rovel:
Hi,

I'm trying to run a basic migration:

```
def change
    update_view :materialized_skill_table_name,
      version: 2,
      revert_to_version: 1,
      materialized: true
  end
```

and it trows this strange error:

```
wrong number of arguments (given 2, expected 1)/usr/local/bundle/gems/highline-1.7.10/lib/highline.rb:620:in `say'
/usr/local/bundle/gems/scenic-1.3.0/lib/scenic/adapters/postgres/index_reapplication.rb:66:in `say'
/usr/local/bundle/gems/scenic-1.3.0/lib/scenic/adapters/postgres/index_reapplication.rb:48:in `try_index_create'
/usr/local/bundle/gems/scenic-1.3.0/lib/scenic/adapters/postgres/index_reapplication.rb:35:in `each'
/usr/local/bundle/gems/scenic-1.3.0/lib/scenic/adapters/postgres/index_reapplication.rb:35:in `on'
/usr/local/bundle/gems/scenic-1.3.0/lib/scenic/adapters/postgres.rb:154:in `update_materialized_view'
/usr/local/bundle/gems/scenic-1.3.0/lib/scenic/statements.rb:86:in `update_view'
/usr/local/bundle/gems/activerecord-5.0.7/lib/active_record/migration.rb:846:in `block in method_missing'
/usr/local/bundle/gems/activerecord-5.0.7/lib/active_record/migration.rb:815:in `block in say_with_time'
/usr/local/bundle/gems/activerecord-5.0.7/lib/active_record/migration.rb:815:in `say_with_time'
/usr/local/bundle/gems/activerecord-5.0.7/lib/active_record/migration.rb:835:in `method_missing'
/app/db/migrate/20170315164338_update_materialized_skill_reports_to_version_2.rb:3:in `change'
/usr/local/bundle/gems/activerecord-5.0.7/lib/active_record/migration.rb:789:in `exec_migration'
/usr/local/bundle/gems/activerecord-5.0.7/lib/active_record/migration.rb:773:in `block (2 levels) in migrate'
/usr/local/bundle/gems/activerecord-5.0.7/lib/active_record/migration.rb:772:in `block in migrate'
[...]

PR #328, opened 2021-04-08
Fix update_view statement from crashing with wrong method access.

rapito:
## Fixes bug and crash with update_view statement

Use `ActiveRecord::Migration` instance instead of class when creatin an `IndexReplication` object to prevent wrong method 
access when using `IndexReplication#speaker`.

## Why?
In short, using `update_view` in a migration crashes because `scenic` tries to do `IndexReplication#speaker#say` on the wrong object.

This happens because the `speaker` object is expected to be an **instance** of `ActiveRecord::Migration` but the `initializer` of `IndexReplication` actually assigns the **class** insteand, so it ends up trying to do `ActiveRecord::Migration::say` instead of `ActiveRecord::Migration#say`.

More info can be found on these comments: 
- https://github.com/scenic-views/scenic/issues/253#issuecomment-698306195
- https://github.com/scenic-views/scenic/issues/253#issuecomment-815184341

## Changes 

+ Make `speaker` default to a `new` **class instance** instead of just the **class**.

## Fixes https://github.com/scenic-views/scenic/issues/253

### DIFF
--- a/lib/scenic/adapters/postgres/index_reapplication.rb
+++ b/lib/scenic/adapters/postgres/index_reapplication.rb
@@ -13,7 +13,7 @@ module Scenic
         # @param connection [Connection] The connection to execute SQL against.
         # @param speaker [#say] (ActiveRecord::Migration) The object used for
         #   logging the results of reapplying indexes.
-        def initialize(connection:, speaker: ActiveRecord::Migration)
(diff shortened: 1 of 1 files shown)


rapito:
Commenting as a friendly bump!

davidalejandroaguilar:
Bumping this PR. Otherwise, a simple view can't be rolled back!

PR #333, opened 2021-07-07
Fix #331: keywords are correctly passed in Ruby 2 and 3

### LINKED ISSUE #331: Reverting breaks on Ruby 3 due to keyword changes, solution is relatively easy
Hi 👋 so the way the command recorder methods record their result, cause the migrations to error when reverting a migration on Ruby 3. Eg here is a method that is incorrect:

https://github.com/scenic-views/scenic/blob/v1.5.4/lib/scenic/command_recorder.rb#L14-L16

```ruby
module Scenic
  module CommandRecorder
    # ...
    def update_view(*args)
      record(:update_view, args)
    end
  end
end
```

After a lot of research (eg @eregon's blog [here](https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html)), I've come to the conclusion that the way to fix this across all the relevant Ruby versions is to conditionally use `ruby2_keywords`:

```diff
module Scenic
  module CommandRecorder
    # ...
    def update_view(*args)
      record(:update_view, args)
    end
+   ruby2_keywords : update_view if respond_to?(:ruby2_keywords, true)
  end
end
```

There's a lot of additional detail in ([this](https://github.com/rails/rails/issues/42653)) bug in ActiveRecord, where I initially reported it.

But you can run the test suite on 2.7 with `env RUBYOPT=-W:deprecated rake` and it will emit warnings where this error occurs. In Ruby 3 , each of those warnings actually explodes. It's pretty confusing to trace them back to Scenic's definition 

env RUBYOPT=-W:deprecated rake

Or you can run it on Ruby 3, it will actually fail in those spots 😜

image

JoshCheek:
* Added Ruby 3.0 to the build matrix
* Skipped Ruby 3 on Rails 5.2 since I don't think they went back and made old Rails work with new Ruby
* Omitted Ruby 2.6 because when I ran it agianst master, it errored saying:
  > activesupport-7.0.0.alpha requires ruby version >= 2.7.0, which is incompatible with the current version, ruby 2.6.3p62`

For more context on why this is the solution, see:

* https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html
* https://github.com/scenic-views/scenic/issues/331
* https://github.com/rails/rails/issues/42653

---

I don't have permission to run CI on this repo, so I ran it on my fork ([here](https://github.com/JoshCheek/scenic/pull/1)), and it [passes](https://github.com/JoshCheek/scenic/pull/1/checks?check_run_id=3005171796):

image

### DIFF
--- a/lib/scenic/command_recorder/statement_arguments.rb
+++ b/lib/scenic/command_recorder/statement_arguments.rb
@@ -36,15 +36,26 @@ module Scenic
         @options ||= @args[1] || {}
       end
 
-      def options_for_revert
-        options.clone.tap do |revert_options|
-          revert_options[:version] = revert_to_version
-          revert_options.delete(:revert_to_version)
+      if Hash.respond_to? :ruby2_keywords_hash
+        def keyword_hash(hash)
+          Hash.ruby2_keywords_hash hash
+        end
+      else
+        def keyword_hash(hash)
+          hash
         end
       end
 
+      def options_for_revert
+        opts = options.clone.tap do |revert_options|
+          revert_options[:version] = revert_to_version
+          revert_options.delete(:revert_to_version)
+        end
+        keyword_hash opts
+      end
+
       def options_without_version
-        options.except(:version)
+        keyword_hash options.except(:version)
       end
     end
   end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,13 +13,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["2.7","2.4"]
+        ruby: ["3.0", "2.7", "2.4"]
         rails: ["5.2", "6.0", "master"]
         exclude:
           - ruby: "2.4"
             rails: "6.0"
           - ruby: "2.4"
             rails: "master"
+          - ruby: "3.0"
+            rails: "5.2"
         include:
           - rails: "master"
             continue-on-error: true
--- a/lib/scenic/command_recorder.rb
+++ b/lib/scenic/command_recorder.rb
@@ -6,18 +6,22 @@ module Scenic
     def create_view(*args)
       record(:create_view, args)
     end
+    ruby2_keywords :create_view if respond_to?(:ruby2_keywords, true)
 
     def drop_view(*args)
       record(:drop_view, args)
     end
+    ruby2_keywords :drop_view if respond_to?(:ruby2_keywords, true)
 
     def update_view(*args)
       record(:update_view, args)
     end
+    ruby2_keywords :update_view if respond_to?(:ruby2_keywords, true)
 
     def replace_view(*args)
       record(:replace_view, args)
     end
+    ruby2_keywords :replace_view if respond_to?(:ruby2_keywords, true)
 
     def invert_create_view(args)
       drop_view_args = StatementArguments.new(args).remove_version.to_a


lib/scenic/command_recorder/statement_arguments.rb
```diff
@@ -36,15 +36,26 @@ def options
         @options ||= @args[1] || {}
       end
 
+      if Hash.respond_to? :ruby2_keywords_hash
+        def keyword_hash(hash)
+          Hash.ruby2_keywords_hash hash
```
eregon, line 41: This should only be used if there is no other way essentially:
```
------------------------------------------------------------------------
  Hash.ruby2_keywords_hash(hash) -> hash

------------------------------------------------------------------------

Duplicates a given hash and adds a ruby2_keywords flag. This method is
not for casual use; debugging, researching, and some truly necessary
cases like deserialization of arguments.
```

Could whatever consumes the `options` from this class use them with `**options` when calling a method with it?
I think that should be sufficient.

OTOH, I can see preserving the ruby2_keywords state of the Hash also makes sense.

JoshCheek: > Could whatever consumes the options from this class use them with **options when calling a method with it?
I think that should be sufficient.

Not sure I understand. ActiveRecord's CommandRecorder ultimately passes them back in, which is why it keeps looking like ActiveRecord is broken. They get passed in to the methods in `Scenic::Statements`, mixed into... IDK, the connection or something, eg here: https://github.com/scenic-views/scenic/blob/v1.5.4/lib/scenic/statements.rb#L25

If you're asking if those methods could take `**options`, then yeah, if you're asking whether the caller (ActiveRecord) can change, then that's less likely, and it's not clear to me that's the right thing to do (b/c right now, ActiveRecord is just taking what you give it and giving it right back)... well... here, it gives it back to you, and asks you to give it the inverse invocation, which is what this code is doing.

I mean, I suppose you could mutate the original hash, as long as you know no one is expecting it to still represent the original invocation. But that feels super iffy.

JoshCheek: Part of the problem is that this keywordness isn't copied in cloning/duping:

```ruby
ruby2_keywords def a(*args)
  Hash.ruby2_keywords_hash? args[-1]        # => true
  Hash.ruby2_keywords_hash? args[-1].clone  # => false
  Hash.ruby2_keywords_hash? args[-1].dup    # => false
end
a b: 1
```

derekprior:
I have to admit I'm a bit out of the loop on this one, but whenever we find ourselves adding shims like these I start to wonder whether it's at all appropriate to drop the older version from support all-together. Are we there yet with Ruby 3? Thoughts, @calebthompson?

lib/scenic/command_recorder/statement_arguments.rb
eregon: It's hard for me to tell without investigating more. What would be useful for me to understand it better is to see the caller and callee code of where the flag gets lost (can be found out with `Hash.ruby2_keyword_hash?(h)`), and to see the callers of the `options` method from this class, and what it does with it.

derekprior:
Reworked a tiny bit but merged in https://github.com/scenic-views/scenic/pull/341. Thanks @JoshCheek and sorry about the delay.

JoshCheek:
> Reworked a tiny bit but merged in #341. Thanks @JoshCheek and sorry about the delay.

Awesome, I just tested it and it works for me 👍 ty ❤️

PR #340, opened 2021-12-14
Add Rails 6.1 to the testing matrix

derekprior:


.github/workflows/ci.yml
```diff
@@ -13,11 +13,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["2.7","2.4"]
-        rails: ["5.2", "6.0", "master"]
+        ruby: ["2.7", "2.4"]
```
calebhearth, line 16: I propose
```suggestion
        ruby: ["ruby-head", "3.0", "2.7"]
```

calebhearth: or master or whatever it is for ruby

calebhearth:
You'll need to skip Rails 6+ on Ruby 2.7

derekprior:
Having bigger problems at the moment -- the tests are broken on 6.1. I think I have a guess as to why but I suspect my fix might break older versions...

PR #341, opened 2021-12-15
Fix issue with keyword arguments under Ruby 3

### LINKED ISSUE #331: Reverting breaks on Ruby 3 due to keyword changes, solution is relatively easy

derekprior:
Ruby 3 changed the handling of keyword arguments in a way that broke our
liberal handling of arguments in the command recorder. Using
`ruby2_keywords` helps us fix this while maintaining compatibility with
older (pre 2.7) versions of Ruby.

In the next major version of Scenic, we should drop support for rubies
older than either 2.7 or 3.0 to avoid carrying this fix forward.

Fixes #331. For more context on why this is the solution, see:

* https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html
* https://github.com/rails/rails/issues/42653

### DIFF
--- a/lib/scenic/command_recorder/statement_arguments.rb
+++ b/lib/scenic/command_recorder/statement_arguments.rb
@@ -36,15 +36,25 @@ module Scenic
         @options ||= @args[1] || {}
       end
 
-      def options_for_revert
-        options.clone.tap do |revert_options|
-          revert_options[:version] = revert_to_version
-          revert_options.delete(:revert_to_version)
+      def keyword_hash(hash)
+        if Hash.respond_to? :ruby2_keywords_hash
+          Hash.ruby2_keywords_hash(hash)
+        else
+          hash
         end
       end
 
+      def options_for_revert
+        opts = options.clone.tap do |revert_options|
+          revert_options[:version] = revert_to_version
+          revert_options.delete(:revert_to_version)
+        end
+
+        keyword_hash(opts)
+      end
+
       def options_without_version
-        options.except(:version)
+        keyword_hash(options.except(:version))
       end
     end
   end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["2.7"]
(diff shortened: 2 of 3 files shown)


lib/scenic/command_recorder/statement_arguments.rb
```diff
@@ -36,15 +36,25 @@ def options
         @options ||= @args[1] || {}
       end
 
+      def keyword_hash(hash)
+        if Hash.respond_to? :ruby2_keywords_hash
```
derekprior, line 40: @JoshCheek's original version conditionally defined this method, which means we don't have to call `respond_to?` every time call `keyword_hash` but `def`ing the method once is more aesthetically pleasing and overall less complex. Given that there is no performance concern with migrations, I think this is the right tradeoff.

PR #344, opened 2021-12-16
Handle ActiveRecord table name prefix and suffix

### LINKED ISSUE #295: Table name prefix (and suffix) are not entirely working
If your application has configured a table name prefix, using the usual `config.active_record.table_name_prefix = 'feds_api_'` method in application.rb, Scenic will properly resolve that name throughout its chain, except when it comes time to locate the SQL file containing the view query.

```sh
No such file or directory @ rb_sysopen - /home/waldavis/projects/feds-api/db/views/feds_api_foos_v01.sql
```

This is an accurate warning, because the generated SQL file from `rails g scenic:view foo` is really named `foos_v01.sql`.

It's important to note that whenever prefixes are used in Rails, you never refer to them in the migration tooling at all. That is to say, if your prefix is `foo_` and you want to create a migration for `bars`, you just use the command `rails g migration create_bars field, field2, field3...` and the resulting migration will not refer to `foo_bars` internally, either, but the generated table will be named `foo_bars`.

So back to our prefixes and Scenic -- this really gets ugly if you naively fix the problem by renaming the generated SQL file so the migration can find it. This works, but only for version 1. If you follow the directions to update, which is to issue the same command you used to create the first (or Nth) version: `rails g scenic:view foo`. Now you get the error that you are going to replace the original migration. Since Scenic uses the names of the files to derive its versioning, it thinks you are trying to create the view again for the first time, and not finding any `foos_v01.sql` it naturally tries to create one, along with the initial `[timestamp]_create_foos.rb`, which already exists.

Finally, if you try to work around the system for the upgrades, by using the command `rails g scenic:view feds_api_foo`, then everything works as documented, except that the generated migration file reads `update_view :feds_api_foo, version: 2, revert_to_version: 1` internally, and then fails to run, complaining that it cannot find a file named `feds_api_feds_api_foos_v02.sql` in the views directory. Again, this is accurate, as the file is named with only one instance of the prefix.

It seems to me that the simplest possible solution would be to ensure that the file path resolution during the migration ignores the prefix. I've fought with this problem in another gem before, and I know that "simple" is a very loaded word here. You're either going to need to do that, or ensure that within your gem, the files are written out with their prefix attached. Either one of those will get around the problem.

Everything I've written here applies to the suffix as well as the prefix.

derekprior:
ActiveRecord allows you to set `table_name_prefix` and `table_name_suffix` both via configuration (which ultimately sets the options on `ActiveRecord::Base`) and in any subclass of ActiveRecord. When set in configuration or directly on `ActiveRecord::Base`, rails migrations and the schema dumper make the setting transparent to migrations, `schema.rb`, etc.

For example, if I run the following:

```
ActiveRecord::Base.table_name_prefix = "api_"

class CreateUsers < ActiveRecord::Migration[7.0]
  def change
    create_table :users
  end
end
```

then the table that is created in the database is `api_users`, but `schema.rb` calls the table `users`. If the user runs `rake db:schema:load` then the table is created as `api_users` once again. Setting the prefix or suffix in any way other than in configuration or directly on `ActiveRecord::Base` does not influence schema generation at all.

Prior to this change, Scenic didn't handle this situation at all. If you had a prefix set and tried to run `create_view :searches` Scenic would fail looking for the file `app_searches_v01.sql`. It turns out, rails migrations mutate the arguments to most every method call automatically. See: https://github.com/rails/rails/blob/4607108f91881cd1c24285dca63dc8e0f3f8a4f1/activerecord/lib/active_record/migration.rb#L917-L933

To get around this, I added `Scenic::UnaffixedName` so we could reverse this process. This is essentially a copy of what Rails itself does in its own schema dumper here: https://github.com/rails/rails/blob/4607108f91881cd1c24285dca63dc8e0f3f8a4f1/activerecord/lib/active_record/schema_dumper.rb#L295-L299

By unaffixing the view name in two places, we're able to fully support table name prefix and suffix. This was probably more investigative work than was warranted for this edge case feature, but I dove in thinking it was going to be easy and now here we are...

Fixes #295


walterdavis:
I think this looks very solid. Given the merry path that Rails takes in this area, I can't think of another way to get the solution. I will give this a try on a project and let you know how it goes.

derekprior:
One thing to note is that you'll still need to refer to the affixed tables names in your view SQL. Scenic can't help you there.

PR #346, opened 2021-12-17
Fix serialization of views with backslashes

derekprior:
Views that contain backslashed (e.g. those that use regular expressions)
were not being properly serialized by the schema dumper. Escaping the
`\` characters in the definition allows them to appear correctly in the
string generated by the Ruby heredoc.


derekprior:
Thanks @bensheldon and sorry for the delay!

bensheldon:
AWESOME! ❤️

PR #349, opened 2022-01-12
Exclude pg_stat_statements_info

### LINKED ISSUE #348: "pg_stat_statements_info" already exists when running rails db:test:load
I ran a scenic migration to create a materialized view and it adds another `create_view` in my `schema.db`

```ruby
create_view "pg_stat_statements_info", sql_definition: <<-SQL
      SELECT pg_stat_statements_info.dealloc,
      pg_stat_statements_info.stats_reset
     FROM pg_stat_statements_info() pg_stat_statements_info(dealloc, stats_reset);
  SQL
```

When running `db:test:load`, I get the following error:

```bash
Caused by:
PG::DuplicateTable: ERROR:  relation "pg_stat_statements_info" already exists
```

Is there any config I might be missing? 🤔

calebhearth:
Postgres 14 [introduced] an internal view `pg_stat_statements_info`, which
we need to exclude from our dumper because including it causes loading
the schema to fail:

[introduced]: <https://www.postgresql.org/docs/14/release-14.html#:~:text=Add%20pg_stat_statements_info%20system%20view%20to%20show%20pg_stat_statements%20activity> "Postgres 14 release notes stating 'Add pg_stat_statements_info system view to show pg_stat_statements activity'"

    PG::DuplicateTable: ERROR:  relation "pg_stat_statements_info" already exists

To test this, enable pg_stat_statements in the dummy database. This is a
reasonable addition as the majority of rails apps are using this
extension and it's cleaner and less error-prone than adding a fake view
of the same name in a specific test for this case.

To support this view, we're currently specifically excluding it in the
query used in the Postgres adapter to load views. Rather then excluding
it by name, we could do an extension of our current exclusion for
extension-owned views, `AND c.relname NOT IN (SELECT extname FROM
pg_extension)`
 by excluding any views with an extension name included
rather than exact matches. I'm not convinced that this is more robust
than excluding exact names has been shown to be by this bug.

Fixes #348

### DIFF
--- a/spec/dummy/db/schema.rb
+++ b/spec/dummy/db/schema.rb
@@ -1,0 +1,19 @@
+# This file is auto-generated from the current state of the database. Instead
+# of editing this file, please use the migrations feature of Active Record to
+# incrementally modify your database, and then regenerate this schema definition.
+#
+# This file is the source Rails uses to define your schema when running `bin/rails
+# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
+# be faster and is potentially less error prone than running all of your
+# migrations from scratch. Old migrations may fail to apply correctly if those
+# migrations use external dependencies or application code.
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 2022_01_12_154220) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "pg_stat_statements"
+  enable_extension "plpgsql"
+
+end
--- a/spec/scenic/schema_dumper_spec.rb
+++ b/spec/scenic/schema_dumper_spec.rb
@@ -95,7 +95,7 @@ describe Scenic::SchemaDumper, :db do
     output = stream.string
 
     expect(output).to include 'create_view "searches"'
-    expect(output).not_to include "ar_internal_metadata"
+    expect(output).not_to include "pg_stat_statements_info"
     expect(output).not_to include "schema_migrations"
   end
 
--- a/lib/scenic/adapters/postgres/views.rb
+++ b/lib/scenic/adapters/postgres/views.rb
@@ -34,6 +34,7 @@ module Scenic
             WHERE
               c.relkind IN ('m', 'v')
               AND c.relname NOT IN (SELECT extname FROM pg_extension)
+              AND c.relname != 'pg_stat_statements_info'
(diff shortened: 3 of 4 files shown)


lib/scenic/schema_dumper.rb
```diff
@@ -30,7 +30,11 @@ def dumpable_views_in_database
     unless ActiveRecord::SchemaDumper.private_instance_methods(false).include?(:ignored?)
       # This method will be present in Rails 4.2.0 and can be removed then.
       def ignored?(table_name)
-        ["schema_migrations", ignore_tables].flatten.any? do |ignored|
+        [
+          "schema_migrations",
+          "pg_stat_statements_info",
```
derekprior, line 35: I'm not sure we should modify this method. It's a direct copy of the method that existed pre Rails 4.2.0. We're covered by the change in the query for any version of Rails.

spec/scenic/schema_dumper_spec.rb
```diff
@@ -95,7 +95,7 @@ class SearchInAHaystack < ActiveRecord::Base
     output = stream.string
 
     expect(output).to include 'create_view "searches"'
-    expect(output).not_to include "ar_internal_metadata"
+    # expect(output).not_to include "pg_stat_statements_info"
```
derekprior, line 98: ❓

spec/dummy/db/schema.rb
```diff
@@ -0,0 +1,25 @@
...
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 2022_01_12_154220) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "pg_stat_statements"
+  enable_extension "plpgsql"
+
+
+  create_view "pg_stat_statements_info", sql_definition: <<-SQL
```
derekprior, line 20: Isn't this the bug we're trying to fix?

calebhearth: Oh shit.

calebhearth: Yeah that's what happened. Confirmed with `rake dummy:db:migrate` and `rake dummy:db:{drop,create,migrate}`.

derekprior (approved):

melsatar:
Can anyone share the solution for this please, I have updated the gem but the same is still happening.

PR #360, opened 2022-04-20
Fix ERB deprecation warning

aliismayilov:
example warning:
```
lib/generators/scenic/model/model_generator.rb:39: warning: Passing safe_level with the 2nd argument of ERB.new is deprecated. Do not use it, and specify other arguments as keyword arguments.
```

followed Ruby standard library style to make distinction on ERB version check:
https://github.com/ruby/ruby/commit/3406c5d


derekprior:
Thank you!

Issue #334, opened 2021-08-24
Indexes are not replaced after upgrading a materialised view


mattgibson:
Two quirks of our setup might explain this

1. the indexes have custom names
2. We are using a custom schema name rather than `public`

Not had time to find out which yet.

Jamedjo:
Filtering on `t.relname` doesn't work because the namespace is prefixed in the call to `update_view`:

```ruby
pry> Scenic::Adapters::Postgres::Indexes.new(connection: ActiveRecord::Base.connection).on('shop.products')
   (1.9ms)              SELECT
              t.relname as object_name,
              i.relname as index_name,
              pg_get_indexdef(d.indexrelid) AS definition
            FROM pg_class t
            INNER JOIN pg_index d ON t.oid = d.indrelid
            INNER JOIN pg_class i ON d.indexrelid = i.oid
            LEFT JOIN pg_namespace n ON n.oid = i.relnamespace
            WHERE i.relkind = 'i'
              AND d.indisprimary = 'f'
              AND t.relname = 'shop.products'
              AND n.nspname = ANY (current_schemas(false))
            ORDER BY i.relname
=> []
```

If we were to filter just on the actual name we would find the index:

```ruby
pry> Scenic::Adapters::Postgres::Indexes.new(connection: ActiveRecord::Base.connection).on('products')
   (5.8ms)              SELECT
              t.relname as object_name,
              i.relname as index_name,
              pg_get_indexdef(d.indexrelid) AS definition
            FROM pg_class t
            INNER JOIN pg_index d ON t.oid = d.indrelid
            INNER JOIN pg_class i ON d.indexrelid = i.oid
            LEFT JOIN pg_namespace n ON n.oid = i.relnamespace
            WHERE i.relkind = 'i'
              AND d.indisprimary = 'f'
              AND t.relname = 'products'
              AND n.nspname = ANY (current_schemas(false))
            ORDER BY i.relname

=> [#<Scenic::Index:0x0000557cea1d5418
  @definition="CREATE UNIQUE INDEX products_unique_sku_dates ON shop.products USING btree (\"Day\", \"Component\", \"SKU\")",
  @index_name="products_unique_sku_dates",
  @object_name="products">]
```

Should this query be updated split "shop.products", filter on `t.relname = 'products'` and to also filter on [`relnamespace`](https://www.postgresql.org/docs/9.3/catalog-pg-class.html)?

PR #364, opened 2022-07-02
Fix recreating indexes for materialized views within custom schemas

fatkodima:
Fixes https://github.com/scenic-views/scenic/issues/334

When materialized view is created within a custom schema and then we try to update it, the original indexes are not retrieved because of the problem within the used SQL to do that.


fatkodima:
@derekprior I am sorry for pinging, but can you, please, take a look at this?

PR #368, opened 2022-08-17
Add flag --replace_view to migration generator

danhixon-copa:
The previous method to achieve this was by running the generator and editing the migration file. This makes this workflow simpler as well as simplifying the documentation of this feature.

I added a spec and updated the README. Let me know if I need to do anything else.


derekprior (approved):
Looks good. Thank you!

PR #376, opened 2023-01-14
ERB interpolation in definition file, and `version: :latest` as an option

gmhawash:
ERB interpolation is helpful when using ruby constants/methods used inside the definition file.
`version: :latest` allows for instructing scenic to use the latest version of the definition found at that commit.  This is useful when we have to drop views due to a dependency on them and having to recreate them afterward.

### DIFF
--- a/lib/scenic/definition.rb
+++ b/lib/scenic/definition.rb
@@ -7,7 +7,7 @@ module Scenic
     end
 
     def to_sql
-      File.read(full_path).tap do |content|
+      ERB.new(File.read(full_path)).result.tap do |content|
         if content.empty?
           raise "Define view query in #{path} before migrating."
         end
--- a/lib/scenic/definition.rb
+++ b/lib/scenic/definition.rb
@@ -3,7 +3,7 @@ module Scenic
   class Definition
     def initialize(name, version)
       @name = name.to_s
-      @version = version.to_i
+      @version = version
     end
 
     def to_sql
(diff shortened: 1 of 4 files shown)


derekprior:
Thanks @gmhawash. This PR contains two features and, unfortunately, I'm not inclined to merge either, though I'd listen to dissent from @calebhearth and certainly willing to hear more about what the community wants.

## The `:latest` version specifier

Scenic provides reversible, consistent, and idempotent management of database views. The ability to specify `:latest`, while convenient at the time of writing, would be counter to those goals.

When a migration is executed shortly after it's written, it's likely that `:latest` will refer to what you intended. If the migration is run at a later date or if there are multiple migrations introduced to the codebase in short order, then developers are likely to get differing schemas when migrations are run and reversed.

## ERB Interpolation in views

I can see why this would be useful as well, but we've resisted adding it to date for many of the same reasons above. It's a sharp knife that we think people are apt to cut themselves with. There's an argument that we should let people use the sharp knife at their own risk, but then we're the one's that have to handle the follow up issues and PRs that attempt to make the knife safer to use, etc...

calebhearth:
I definitely agree on ERB interpolation. It was something we considered in the initial version of the gem and decided against for the reasons Derek outlined. Notably, the result of the ERB interpolation will not necessarily be the same between different times that the code is executed, so you may get different results after some change to the application or even across different environments.

I've thought through :latest less, but I think I agree with Derek here as well.

PR #379, opened 2023-01-24
Remove extra whitespace from generated update migrations

mrtnin:
The rails generator adds an annoying newline when creating `update_view` or `replace_view` migrations where the ruby is evaluated.


eserra:
+ 1!

PR #385, opened 2023-03-03
add custom paths for scenic views

AlexeyAfrosin:


derekprior:
Thanks for the PR. This has given me some inspiration for how I think we should handle this. I hope to tackle it as part of #402.

As is, this is not mergeable as it takes us further down the single database path.

Issue #386, opened 2023-04-04
make update_view for materialized views create less downtime


Roguelazer:
Right now, `update_materialized_view` results in a `DROP MATERIALIZED VIEW` followed by a `CREATE MATERIALIZED VIEW` with the new version; if the view is large, this can result in substantial downtime, since the drop acquires an ACCESS EXCLUSIVE lock on the view and prevents any queries from running.

Another option would be to do `CREATE MATERIALIZED VIEW {name}_tmp ; ALTER MATERIALIZED VIEW {name} RENAME TO {name}_old ; ALTER MATERIALIZED VIEW {name}_new RENAME TO {name} ; DROP MATERIALIZED VIEW {name}_old`. This results in negligible downtime, but costs twice the disk space. A further improvement could be to optionally run the (potentially expensive, disk-blocking) DROP in a different transaction if you pass an argument like 'i_will_clean_up_myself: true`, but in my experimentation, just doing it this way takes some expensive matview migrations from minutes of downtime to fractions of a second.

I've attached a PR for this behavior.

derekprior:
I really like this idea and have manually done the same many years ago when dealing with a project that used materialized views heavily.

We'll need to be cognizant of things like view and index name length limits. I see there's a PR already associated to this. I likely won't be able t have a look until I'm back from vacation but just wanted to say I'm generally in support of the idea.

PR #387, opened 2023-04-04
Add side_by_side mode to update_materialized_view

Roguelazer:
This adds a `side_by_side` kwarg to the `update_materialized_view` method, which builds the new view alongside the old one and then atomically swaps them to reduce downtime at the cost of increasing disk usage. It is plumbed through to migrations as a hash value for the `materialized` kwarg of `update_view`.

Fixes scenic-views/scenic#386.

### DIFF
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -155,17 +155,30 @@ module Scenic
       # @param no_data [Boolean] Default: false. Set to true to create
       #   materialized view without running the associated query. You will need
       #   to perform a non-concurrent refresh to populate with data.
+      # @param side_by_side [Boolean] Default: false. Set to true to create the new
+      #   version under a different name and atomically swap them, limiting downtime
+      #   at the cost of doubling disk usage
       #
       # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
       #   in use does not support materialized views.
       #
       # @return [void]
-      def update_materialized_view(name, sql_definition, no_data: false)
+      def update_materialized_view(name, sql_definition, no_data: false, side_by_side: false)
         raise_unless_materialized_views_supported
 
         IndexReapplication.new(connection: connection).on(name) do
-          drop_materialized_view(name)
-          create_materialized_view(name, sql_definition, no_data: no_data)
+          if side_by_side
+            session_id = Time.now.to_i
+            new_name = "#{name}_new_#{session_id}"
+            old_name = "#{name}_drop_#{session_id}"
+            create_materialized_view(new_name, sql_definition, no_data: no_data)
+            rename_materialized_view(name, old_name)
+            rename_materialized_view(new_name, name)
+            drop_materialized_view(old_name)
+          else
+            drop_materialized_view(name)
+            create_materialized_view(name, sql_definition, no_data: no_data)
+          end
         end
       end
 
@@ -181,6 +194,19 @@ module Scenic
       def drop_materialized_view(name)
         raise_unless_materialized_views_supported
         execute "DROP MATERIALIZED VIEW #{quote_table_name(name)};"
+      end
+
+      # Renames a materialized view from {name} to {new_name}
+      #
+      # @param name The existing name of the materialized view in the database.
+      # @param new_name The new name to which it should be renamed
+      # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
+      #   in use does not support materialized views.
+      #
+      # @return [void]
+      def rename_materialized_view(name, new_name)
+        raise_unless_materialized_views_supported
+        execute "ALTER MATERIALIZED VIEW #{quote_table_name(name)} RENAME TO #{quote_table_name(new_name)};"
       end
 
       # Refreshes a materialized view from its SQL schema.
--- a/lib/scenic/adapters/postgres.rb
+++ b/lib/scenic/adapters/postgres.rb
@@ -22,6 +22,8 @@ module Scenic
     # The methods are documented here for insight into specifics of how Scenic
     # integrates with Postgres and the responsibilities of {Adapters}.
     class Postgres
+      MAX_IDENTIFIER_LENGTH = 63
+
       # Creates an instance of the Scenic Postgres adapter.
       #
       # This is the default adapter for Scenic. Configuring it via
@@ -137,8 +139,8 @@ module Scenic
 
         execute <<-SQL
   CREATE MATERIALIZED VIEW #{quote_table_name(name)} AS
-  #{sql_definition.rstrip.chomp(';')}
-  #{'WITH NO DATA' if no_data};
+  #{sql_definition.rstrip.chomp(";")}
+  #{"WITH NO DATA" if no_data};
         SQL
       end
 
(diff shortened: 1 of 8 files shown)


Roguelazer:
it's somewhat maddening that this repository doesn't have a working `rubocop.yml` (what is in there is a many-years-old version of rubocop and seems to disagree with how hound is configured).

ferblape:
We also implemented something similar in a project using Scenic, but we had some issues with the indexes of the view you are replacing: you should drop them in the current view before adding the swap one, otherwise when you rename the view you'll have conflicts in the naming.

Roguelazer:
Ah, good point. I could probably modify this to rename all the indices, too…

Roguelazer:
Actually, this won't cause conflicts because the indexes on the new view get created after the swap, and the drop at the end of the swap removes the conflicting indexes.

That being said, building the indexes after the drop means that the indexes are built while holding the access exclusive lock, which probably is not a good idea.

scottjacobsen:
This is what `refresh materialized view concurrently...` does - why not use that? https://www.postgresql.org/docs/current/sql-refreshmaterializedview.html

renchap:
@calebhearth @derekprior Gentle ping on this, do you need anything else for this PR to be merged?

This would really help some Mastodon materialized view changes :)

Roguelazer:
Any changes you'd like me to make to this?

gsinkin:
Just adding another nudge here

Roguelazer:
Anything you'd like me to do to help move this forward?

lib/scenic/adapters/postgres.rb
```diff
@@ -155,17 +155,36 @@ def create_materialized_view(name, sql_definition, no_data: false)
...
+      #   downtime at the cost of doubling disk usage
       #
       # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
       #   in use does not support materialized views.
       #
       # @return [void]
-      def update_materialized_view(name, sql_definition, no_data: false)
+      def update_materialized_view(
+        name, sql_definition, no_data: false, side_by_side: false
+      )
```
derekprior, line 168: I know hound is not going to like this due to line length, but I don't care....

```suggestion
      def update_materialized_view(name, sql_definition, no_data: false, side_by_side: false)
```

lib/scenic/adapters/postgres.rb
```diff
@@ -155,17 +155,36 @@ def create_materialized_view(name, sql_definition, no_data: false)
...
+      #   new version under a different name and atomically swap them, limiting
+      #   downtime at the cost of doubling disk usage
       #
       # @raise [MaterializedViewsNotSupportedError] if the version of Postgres
       #   in use does not support materialized views.
       #
       # @return [void]
-      def update_materialized_view(name, sql_definition, no_data: false)
+      def update_materialized_view(
+        name, sql_definition, no_data: false, side_by_side: false
```
derekprior, line 167: I'm not sure `side_by_side` is the name I'd opt for here but I'm lacking a better suggetion at the moment other than `background` which I don't love either.

In your experience, is `side_by_side` an established name for this process?

lib/scenic/adapters/postgres.rb
```diff
@@ -155,17 +155,36 @@ def create_materialized_view(name, sql_definition, no_data: false)
...
+        if side_by_side
+          session_id = Time.now.to_i
+          new_name = "#{name}_new_#{session_id}"
+          old_name = "#{name}_drop_#{session_id}"
+          IndexReapplication.new(connection: connection).on_side_by_side(
+            name, new_name, session_id
+          ) do
+            create_materialized_view(new_name, sql_definition, no_data: no_data)
+          end
+          rename_materialized_view(name, old_name)
```
derekprior, line 180: Is there a reason not to `drop` here?

lib/scenic/adapters/postgres.rb
```diff
@@ -155,17 +155,36 @@ def create_materialized_view(name, sql_definition, no_data: false)
...
+        name, sql_definition, no_data: false, side_by_side: false
+      )
         raise_unless_materialized_views_supported
 
-        IndexReapplication.new(connection: connection).on(name) do
-          drop_materialized_view(name)
-          create_materialized_view(name, sql_definition, no_data: no_data)
+        if side_by_side
+          session_id = Time.now.to_i
+          new_name = "#{name}_new_#{session_id}"
```
derekprior, line 173: Postgres identifiers are limited to 63 characters. We're removing 15 characters from that allotment already. This will likely be problematic for some view names. What does Rails do when, for example, a generated index name would be too long? We may need to access that functionality (or copy it) here.

lib/scenic/adapters/postgres/index_reapplication.rb
```diff
@@ -35,6 +35,18 @@ def on(name)
           indexes.each(&method(:try_index_create))
         end
 
+        def on_side_by_side(name, new_table_name, temporary_id)
+          indexes = Indexes.new(connection: connection).on(name)
+          indexes.each_with_index do |index, i|
+            old_name = "predrop_index_#{temporary_id}_#{i}"
```
derekprior, line 41: This looks like it cuts 24 characters off the alloted 63 for postgres identifiers (see earlier comment). We may need to think more about how to lessen the chance this is an issue for folks

derekprior (changes requested):
I think there are some issues that we're likely to encounter with this in the wild that require some more thought.

I also would like to encapsulate the functionality a bit differently but I don't have and easily explainable idea at the moment. Best bet may be to tackle the outstanding questions to get to a mergeable idea and then we can refactor after that.

lib/scenic/adapters/postgres.rb
Roguelazer: I think "online" might be a more established name but I was trying to think of something that was very specific about what it did

lib/scenic/adapters/postgres.rb
Roguelazer: IIRC they use a truncated hash of the generated components.

The name only matters for debugging, so I'd be fine with doing that. Something like

```ruby
new_name = "#{name}_new_#{session_id}"
if new_name.size > 63
  # session_id is 10c, _new_ is 5c; leave lots of space free
  new_name = "#{Digest::SHA256.hexdigest(name)[...40]}_new_#{session_id}"
end
```

lib/scenic/adapters/postgres/index_reapplication.rb
Roguelazer: This one doesn't actually have the index name in it (it just looks like `predrop_index_1705535993_1`). I'm pretty sure I did it that way _because_ I hit the index size limit. :-D

lib/scenic/adapters/postgres.rb
Roguelazer: I think I did it this way out of the assumption that we might commit after the rename (but before the drop) and pause to give other threads a chance to finish using that view before we drop it, to reduce the lock impact. I didn't actually build that, of course.

lib/scenic/adapters/postgres.rb
Roguelazer: I'd need to also add some comments to make rubocop be quiet. I don't mind doing so (and actually set `Metrics/LineLength` to 120 on all my projects anyway).

Roguelazer: I guess rubocop isn't automatically run (and the config file is remarkably old; at least <1.7 and broken with current rubocop versions) so I won't worry about it

mjankowski:
Rebasing this after https://github.com/scenic-views/scenic/pull/397 will likely get rid of hound noise in this PR.

derekprior:
#430 is the rebased version of this where work will continue. Would like to get this merged in the next week. Just needs a bit of a cleanup and some work on edge cases.

PR #396, opened 2023-09-29
Add method to check materialized view population

derekprior:
REFRESH MATERIALIZED VIEW CONCURRENTLY doesn't work for a view which is not populated, even it has unique index. This method is useful to perform concurrent refresh if it's possible.

Real use case in Rails will be like this:

* Database is loaded from structure.sql to run specs, but the view isn't populated yet so concurrent refresh is not possible. We want to fallback to normal(non-concurrent) refresh in this case.
* After initial population, views are expected to be refreshed concurrently. It's possible because the view is already populated.


derekprior:
Replaces #249

calebhearth (approved):

PR #397, opened 2023-09-29
Add Standard for Ruby formatting

derekprior:
We had been using Hound CI and a custom RuboCop config to encourage adhering to our style guide. In the last couple of years, Standard has emerged and gained steam as a "set it and forget it" RuboCop configuration for Ruby. Maintaining our own opinionated RuboCop configuration is not something I care to do any longer. This change:

1. Adds Standard
2. Fixes all[^1] of the issues it detects
3. Adds Standard to CI
4. Removes our hound and RuboCop configuration

Adding Standard as an explicit CI step will require compliance for passing builds, which is a departure from our previous setup of only commenting via Hound. I think the time is right for this change given the maturity of auto-fixers and integrations within various Ruby workflows.

[^1]: We continue to allow the use of `eval` in our specs


derekprior:
Hound is still configured as a hook on this repo, which is why it ran on this PR with its default configuration (since I removed our configuration). I have resolved all of the comments and will remove the hound hook if we merge this.

derekprior:
Not 100% sure why I can't get standard to install correctly on CI. Will have to come back to it.

derekprior:
Whelp, got it working but it's not erroring on a violation I intentionally committed...

derekprior:
This is working now, with annotations on failure. 🎉

PR #337, opened 2021-09-24
Add topological sorting for dumped views using TSort

msorc:
This introduces topological sorting (https://ruby-doc.org/stdlib-2.6.8/libdoc/tsort/rdoc/TSort.html#method-i-tsort) for views in dumps which will sort views alphabetically and in case of dependencies between views will use a topological sorting. 
That provides a consistent views ordering in database dumps and prevents of having occasional diffs in schema dumps.

Based on discussion here https://github.com/scenic-views/scenic/issues/263 and code samples https://github.com/scenic-views/scenic/issues/263#issuecomment-679167557 and https://github.com/scenic-views/scenic/issues/263#issuecomment-875009521

PR #398, opened 2023-10-25
Add topological sorting for dumped views via TSort

calebhearth:
Rebase of https://github.com/scenic-views/scenic/pull/337

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,8 +30,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ["2.7", "3.0", "3.1", "3.2"]
-        rails: ["6.1", "7.0"]
+        ruby: ["3.1", "3.2", "3.3"]
+        rails: ["6.1", "7.0", "7.1", "main"]
         continue-on-error: [false]
         exclude:
           - ruby: "3.2"
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,3 @@
+Caleb Hearth <caleb@calebhearth.com> <caleb@calebthompson.io>
+Derek Prior <derekprior@gmail.com> <derek@thoughtbot.com>
+Devon Estes <devon.c.estes@gmail.com> <devoncestes@devons-air.localdomain>
--- a/Gemfile
+++ b/Gemfile
@@ -3,14 +3,13 @@ source "https://rubygems.org"
 # Specify your gem's dependencies in scenic.gemspec
 gemspec
 
-rails_version = ENV.fetch("RAILS_VERSION", "6.1")
+rails_version = ENV.fetch("RAILS_VERSION", "7.1")
 
-rails_constraint = if rails_version == "master"
+rails_constraint = if rails_version == "main"
   {github: "rails/rails"}
 else
   "~> #{rails_version}.0"
 end
 
-gem "rails", rails_constraint
-gem "sprockets", "< 4.0.0"
-gem "pg", "~> 1.1"
+gem "activerecord", rails_constraint
+gem "railties", rails_constraint
--- a/lib/scenic/adapters/postgres/connection.rb
+++ b/lib/scenic/adapters/postgres/connection.rb
@@ -46,6 +46,8 @@ module Scenic
           end
         end
 
+        def with_connection(&) = connection_pool.with_connection(&)
+
         private
 
         def undecorated_connection
--- a/lib/scenic/schema_dumper.rb
+++ b/lib/scenic/schema_dumper.rb
@@ -67,8 +67,14 @@ module Scenic
       views_hash = TSortableHash.new
 
       ::Scenic.database.execute(DEPENDENT_SQL).each do |relation|
-        source_v = relation["source_table"]
-        dependent = relation["dependent_view"]
+        source_v = [
+          relation["source_schema"],
+          relation["source_table"],
+        ].compact.join(".")
+        dependent = [
+          relation["dependent_schema"],
+          relation["dependent_view"],
+        ].compact.join(".")
         views_hash[dependent] ||= []
         views_hash[source_v] ||= []
         views_hash[dependent] << source_v
--- a/lib/scenic/schema_dumper.rb
+++ b/lib/scenic/schema_dumper.rb
@@ -8,8 +8,8 @@ module Scenic
       include TSort
 
       alias_method :tsort_each_node, :each_key
-      def tsort_each_child(node, &block)
-        fetch(node).each(&block)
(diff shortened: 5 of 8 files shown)


edwardloveall:
Hello! We (me and @thorncp) have been trying this code, but it hasn't worked for us. Here's exactly what we tried:

* pin scenic to `1431c49ec6d6063212aa718db73cc6158faa1461`
* create view `as`
* migrate
* This added `as` view at the top of the views in db/schema (alphabetical sort)
* create view `bs`
* migrate
* added `bs` view after `as` in the db/schema (alphabetical sort)
* `replace_view` `as` with `select * from bs` to generate dependent order
  * We also tried it with `update_view` and got the same outcome
* migrate
* `as` replaced the view and _did not_ move it below `bs`

At this point, our schema can't be loaded into the database, because it will try to create `as` using `bs` but `bs` does not yet exist.

```sh
rails db:schema:load

ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "bs" does not exist
LINE 2:      FROM bs;
                  ^
/path/to/db/schema.rb:4416:in `block in <main>'
/path/to/db/schema.rb:13:in `<main>'

Caused by:
PG::UndefinedTable: ERROR:  relation "bs" does not exist
LINE 2:      FROM bs;
                  ^
```

We also tried this with our existing views. Before we dumped the schema they were in a valid, topological order. After we dumped, they were in an invalid non-topological order.

It this expected? As it stands, our app would break if this got merged.

derekprior:
This is not expected. The idea of this PR is to provide both correct ordering (which we have today) and stable ordering (which we do not).

Your manual steps are a good test case for us to consider including in our suite today (that `bs` must come before `as`) to prevent regression. I'm a bit surprised we don't have something like that already though...

edwardloveall:
Thanks for the confirmation.

FWIW, I just tried to make a test for this and it passes:

```rb
it "sorts dependency order over alphabetical order" do
  conn.create_view "as", sql_definition: <<-SQL
    SELECT 1;
  SQL
  conn.create_view "bs", sql_definition: <<-SQL
    SELECT 2;
  SQL
  conn.update_view "as", sql_definition: <<-SQL
    SELECT * FROM bs;
  SQL
  
  expect(views).to eq(%w[bs as])
end
```

I also tried `conn.replace_view` but that doesn't take a `sql_definition` and I didn't research further.

Maybe it's our setup somehow? Maybe there's some untested behavior if you migrate in between creating and updating dependent views?

edwardloveall:
Update: I tried this on a brand new app and it _did_ put `bs` before `as`. So I'm thinking it has something to do with our particular app. Hopefully I'll get some more info for you.

edwardloveall:
We (@heatherbooker and I) figured it out today. Our database views are not in the `public` schema, they exist in a different schema set via the `config/database.yml` that looks something like this:

```yaml
production:
  database: app_production
  schema_search_path: "abc"
```

[The SQL that generates the list of dependent views](https://github.com/scenic-views/scenic/pull/398/files#diff-20b7aab2911168a07f8b99600120a165c4c08b79aa1cd6b932e152fda4eb623dR18-R31) does not include the schema name, but [the code to generate view names](https://github.com/scenic-views/scenic/blob/1431c49ec6d6063212aa718db73cc6158faa1461/lib/scenic/adapters/postgres/views.rb#L46-L50) _does_ include the non-`public` schema name.

This breaks the [logic that prepares the views to be TSorted](https://github.com/scenic-views/scenic/pull/398/files#diff-20b7aab2911168a07f8b99600120a165c4c08b79aa1cd6b932e152fda4eb623dR73-R79).

We updated the select portion of the SQL and now our example case passes and puts `as` below `bs`:

```sql
SELECT DISTINCT
    dependent_ns.nspname AS dependent_schema
  , dependent_view.relname
  , (dependent_ns.nspname || '.' || dependent_view.relname) AS dependent_view
  , source_ns.nspname AS source_schema
  , (source_ns.nspname || '.' || source_table.relname) AS source_table
-- rest of the SQL
```

This will need some updating because it prints `public.view_name` which does not match the logic in scenic so it will break existing tests and functionality, but it should be a pretty easy fix.

calebhearth:
At my jobby job we've been using this since 10/26 and have made several updates to views (though none that change the dependency structure) and have _not_ seen any unnecessary reordering of views, so I'm inclined after ~2.5 months to call this a win.

@edwardloveall I wonder if #401 helps with your use case at all? Also if you want to write up a PR to add specs that exercise view ordering I'd appreciate it.

derekprior:
It's good enough for me. Merge when ready.

When we had talked about this problem originally (years ago), we said we would make this a "breaking change" for 2.0 since it would likely change everyone's generated structure.sql once. Do you still feel that's appropriate?

edwardloveall:
> @edwardloveall I wonder if #401 helps with your use case at all?

It does, but only in combination with the changes here.

> Also if you want to write up a PR to add specs that exercise view ordering I'd appreciate it.

Done: https://github.com/scenic-views/scenic/pull/403

calebhearth:
I'm seeing a failure on one of the specs here after @edwardloveall's PR adding sort order specs was merged. I'm fairly sure that the spec is correct and I'm not sure why the failure is happening.

edwardloveall:
> I'm seeing a failure on one of the specs here after @edwardloveall's PR adding sort order specs was merged. I'm fairly sure that the spec is correct and I'm not sure why the failure is happening.

Looking at the branch, it's similar to my [previous comment](https://github.com/scenic-views/scenic/pull/398#issuecomment-1819830928): the SQL that generates the dependent views does not contain the schema, but Scenic's views do. `tsorted_views` then sorts the non-schema-qualified views correctly and ignores the schema-qualified views. Then the non-schema-qualified views get filtered out.

Here's one possible fix:

```diff
  def tsorted_views(views_names)
    views_hash = TSortableHash.new
  
    ::Scenic.database.execute(DEPENDENT_SQL).each do |relation|
-     source_v = relation["source_table"]
-     dependent = relation["dependent_view"]
+     source_v = [relation["source_schema"], relation["source_table"]].join(".")
+     dependent = [relation["dependent_schema"], relation["dependent_view"]].join(".")
      views_hash[dependent] ||= []
      views_hash[source_v] ||= []
      views_hash[dependent] << source_v
      views_names.delete(source_v)
      views_names.delete(dependent)
    end
  
    # after dependencies, there might be some views left
    # that don't have any dependencies
    views_names.sort.each { |v| views_hash[v] = [] }
    binding.irb
  
    views_hash.tsort
  end
```

calebhearth:
Superseded by https://github.com/scenic-views/scenic/pull/416

PR #403, opened 2024-01-22
Test for schema order of non-public-schema views

edwardloveall:
Upcoming changes like [topological sorting][1] have the potential to write views to `db/schema.rb` in an order that does not respect dependencies. This adds a test to ensure that views in all schemas are both included, and in the correct dependency order.

[1]: https://github.com/scenic-views/scenic/pull/398


calebhearth:
Thanks @edwardloveall!

PR #410, opened 2024-03-22
README: Update example ActiveRecord model

franzliedke:
In order to use the model to `find` individual records, the primary key must be explicitly defined.

At least that's my experience with PostgreSQL.

Also, see https://arjanvandergaag.nl/blog/view-backed-activerecord-models.html


derekprior:
Thank you. @calebhearth do you think we should include this in the generated model? https://github.com/scenic-views/scenic/blob/main/lib/generators/scenic/model/templates/model.erb

PR #439, opened 2025-02-10
Clarify where to use the cascade option for materialized views

Joeventures:
I was reading through the section about materialized views and thought the cascade option was helpful, but it wasn't clear where it should be used.


calebhearth:
Thank you!

PR #452, opened 2025-10-28
ensure multi-schema dumping in Rails 8.1 doesn't dump the same objects for all schemas

agrobbin:
In Rails 8.1, `bin/rails db:schema:dump` now knows how to dump multiple schemas (rails/rails#50020). To ensure we don't dump the same functions / triggers for each schema, we need to stop memoizing `Scenic.database.views`.

This should be fully backwards compatible to previous Rails versions, as we still only call each of those methods once per schema dumping invocation, it just now works if schema dumping is invoked multiple times.

### DIFF
--- a/lib/scenic/schema_dumper.rb
+++ b/lib/scenic/schema_dumper.rb
@@ -9,6 +9,10 @@ module Scenic
     end
 
     def views(stream)
+      dumpable_views_in_database = Scenic.database.views.reject do |view|
+        ignored?(view.name)
+      end
+
       if dumpable_views_in_database.any?
         stream.puts
       end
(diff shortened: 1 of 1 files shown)


robinator:
This worked great for our app 💯 !

calebhearth:
Thank you both!

timon:
Folks, would you mind pushing out a rubygem release with this change?
We recently switched to Rails 8.1 and I was surprised seeing `create_views` disappearing from schema file

timon:
A heads up, though, schema dumper dumps indexes for materialized views incorrectly:
```ruby
  add_index public."vw_attributions", ["type"], name: "index_vw_attributions_on_type" # not "public.vw_attributions"
````

Update: filed an issue under rails/rails#56301

denysivanov:
@calebhearth, would you mind pushing out a rubygem release with this change?

sk-:
@calebhearth are there any plans to release this?

Issue #312, opened 2020-12-01
Is there a way to turn off pluralization of the view name?


MadBomber:
I'm testing scenic to see how it scratches my itch.  The first unexpected thing I found is that when I give the scenicview the name of the view I want it tells me no you don't want that view name you want one that is pluralized.  My standard naming convention is that the view name ends like "*_view" not as scenic demands *_views

Have you got an option in the generator to turn off pluralization?

If not, when I bypass the generator and create my own sql and migration files will scenic play nice with *_view named views?

MadBomber:
Bypassing the generator and using my own naming convention does not seem to be a problem.

The primary benefit to me is keeping a record of the changes to the view configuration over time and the ability to rollback should the need arise.

derekprior:
We pluralize similar to rails models/migrations. In fact, under the covers we rely on some of the same generators.

flanger001:
Apologies for necro-posting here but this issue seemed like the most correct place to put this:

I have an app that has `config.active_record.pluralize_table_names = false` in `application.rb` and I feel like the least-surprise behavior would be that Scenic would respect that setting, but as of version 1.9, it does not:

```shell
$ ./bin/rails g scenic:view foo
      create  db/views/foos_v01.sql
      create  db/migrate/20260311211130_create_foos.rb
```

`config.active_record.pluralize_table_names` has been around since the early days of Rails, so there is some precedent here. Would you accept a PR for this?

derekprior:
@flanger001 Yes, probably. So long as it plays well with the rest of the generators.

flanger001:
@derekprior Once again, apologies for the necro-post but what do you think of the PR I opened about this?

flanger001:
All right. @calebhearth maybe I can get your eyes on this?

PR #459, opened 2026-03-25
Respect pluralize_table_names option when creating views

flanger001:
References #312

As a preamble, I have followed all the steps in `CONTRIBUTING.md` and have attempted to change the smallest number of lines of code possible.

This PR allows Scenic to respect the `pluralize_table_names` option when creating views. Formerly, if `pluralize_table_names` was set to `false`, Scenic would still create a pluralized view and migration file, e.g.:

```rb
# config/application.rb

Rails.application.configure do
  # snip
  config.active_record.pluralize_table_names = false
end
```

The generator command:

```shell
./bin/rails g scenic:view search
``` 

Outputs:

- `db/migrate/TIMESTAMP_create_searches.rb` 
- `db/views/searches_v01.sql`.

Now, with the aforementioned config, this is our output:

- `db/migrate/TIMESTAMP_create_search.rb` 
- `db/views/search_v01.sql`.

Updating migrations works as well: `./bin/rails generate scenic:view search` a second time will output:

- `db/migrate/TIMESTAMP_update_search_to_version_2.rb` 
- `db/views/search_v02.sql`.

### DIFF
--- a/lib/generators/scenic/view/templates/db/migrate/create_view.erb
+++ b/lib/generators/scenic/view/templates/db/migrate/create_view.erb
@@ -1,5 +1,5 @@
 class <%= migration_class_name %> < <%= activerecord_migration_class %>
   def change
-    create_view <%= formatted_plural_name %><%= create_view_options %>
+    create_view <%= formatted_name %><%= create_view_options %>
   end
 end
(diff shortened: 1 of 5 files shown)


flanger001:
@derekprior What do you think of this approach?

